PR #20, opened 2022-02-15
pdb.rs: Fix Table RowGroup parsing

Swiftb0y:
The reference documentation contains an error which caused the
parsing to shift. This resulted in the effect that page_flags
and RowGroups could be parsed incorrectly.

see https://github.com/Deep-Symmetry/crate-digger/pull/23


Holzhaus (approved):
LGTM, thanks!

Issue #21, opened 2022-02-15
`Row::parse_album` broken


Swiftb0y:
Leaving here as a TODO so no one forgets. 
Currently `Row::parse_album` uses `Row::parse_string_offset` which expects the offset to be a `u16`, however, in the
case of Album rows, the offset is only a `u8`. So parsing that as a `u16` usually results in bogus values which causes
the parser to break when interpreting data at those offsets as `DeviceSQLString`s.

PR #22, opened 2022-02-16
pdb.rs: fix Row::parse_album string offset

Swiftb0y:
quick fix for #21


Holzhaus (approved):
This might also be refactored into a `parse_string_offset_u8` method, but since it's only used once, this works as well.

Holzhaus:
CI passed, thank you.

PR #24, opened 2022-02-16
pdb.rs: implement proper PageIndex iteration

### LINKED ISSUE #23: `Table::page_indicies` incorrect implementation
Pages are a singly-linked lists. Their order can be completely arbitrary. [Modelling them using a monotonic range](https://github.com/Holzhaus/rekordcrate/blob/main/src/pdb.rs#L136-L143) within a table is incorrect.

Swiftb0y:
The previous code worked under the assumption that a tables pages
had monotonic indicies from the first page to the last page. In
reality, pages are organized as linked lists which don't seem
to have any ordering constraints. closes #23

I'm not super familiar how to work with lifetimes properly so feel free to use this as a template for your own implementation.

### DIFF
--- a/src/bin/rekordcrate-pdb.rs
+++ b/src/bin/rekordcrate-pdb.rs
@@ -17,7 +17,7 @@ fn main() {
 
     for (i, table) in header.tables.iter().enumerate() {
         println!("Table {}: {:?}", i, table.page_type);
-        for page_index in table.page_indices() {
+        for page_index in table.page_indices(&header, &data) {
             let (_, page) = header
                 .page(&data, &page_index)
                 .expect("failed to parse page");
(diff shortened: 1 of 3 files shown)


Swiftb0y:
@Holzhaus what do you think of this? IMO this is essential for more complex DBs to work correctly (otherwise were just iterating non-existent pages and/or missing some). I've used this as a base for all other PRs so far and I've parsed a 955 Track DB correctly (without crashing but still finding all tracks).

Holzhaus:
Yes, in general it is correct, but I still have to do an in-depth review.

Holzhaus:
@Swiftb0y I pushed a new commit that avoids an unnecessary page lookup and simplifies the implementation a bit. Please have a look and let me know what you think.

Swiftb0y:
Yes, very elegant solution. I'm jealous I didn't come up with that myself.

PR #26, opened 2022-02-16
pdb.rs: implement ISRC-string parsing

### LINKED ISSUE #25: Track ISRC strings
Attempting to parse Tracks with an ISRC string break the parser because their format falls out of the norm. See the following for more details: https://djl-analysis.deepsymmetry.org/rekordbox-export-analysis/exports.html#long-utf16le

Swiftb0y:
[Weirdness in the database format](https://djl-analysis.deepsymmetry.org/rekordbox-export-analysis/exports.html#long-utf16le)
requires this hack to parse ISRC strings properly. If this is not
accounted for, the contained string could contain invalid
utf16, causing the parser to crash.
fixes #25

### DIFF
--- a/src/pdb.rs
+++ b/src/pdb.rs
@@ -473,6 +473,9 @@ pub enum DeviceSQLString {
     LongASCII(String),
     /// A long UTF-16 little-endian encoded string (max 32767 characters).
     LongUTF16LE(String),
+    /// A ShortASCII variant that is needed to workaround the special ISRC string
+    /// NOTE: the serialized string is a null-terminated C-String
+    ISRCASCII(String),
 }
 
 impl DeviceSQLString {
@@ -480,7 +483,7 @@ impl DeviceSQLString {
         let (_, length_and_kind) = nom::number::complete::u8(input)?;
         match length_and_kind {
             0x40 => Self::parse_long_ascii(input),
-            0x90 => Self::parse_long_utf16le(input),
+            0x90 => nom::branch::alt((Self::parse_isrc_string, Self::parse_long_utf16le))(input),
             _ => Self::parse_short_ascii(input),
         }
     }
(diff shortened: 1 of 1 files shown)


Holzhaus:
Great, can we add a test for that? I didn't bother with unittests for now, but a smoke test would be nice, i. e. adding a test export that contains such a string to the `data/` directory.

Swiftb0y:
The issue is that the current DB I'm working with was provided by a third party and they didn't want me to share it. So I'm afraid I can't really add a test without having to create a whole entire DB myself. I'd rather just create a unit test for it.

Holzhaus:
> I'd rather just create a unit test for it.

Yeah sure, that works, too. I just thought it would be less work to just drop some files into the `data/` directory and thought it would be unfair to requires writing a unittest when I didn't do that either.

Swiftb0y:
Well, in my case dropping some files would first involve painfully creating them in a VM, so I prefer the second option.

Swiftb0y:
Converted to draft because it requires further testing. ISRC strings can be valid UTF16 even though they should not be interpreted that way.

Swiftb0y:
Simply trying the stricter ISRC parser before the UTF16 parser seems to do the trick quite well.

pbeart:
Hey, I think this pdb may have come from me (here's a subset of my playlists to check 
image
), if it is me then I'm happy for it to be included as a test case

Swiftb0y:
Hey there @pbeart 
Indeed, I have received a DB of yours via someone else.
Thanks for your offer. I leave the decision to include other DBs up to @Holzhaus
Your database would be quite useful.

Holzhaus:
> I leave the decision to include other DBs up to @Holzhaus

Sure, just create a new subdirectory in `data/`, put the files in there and open a PR. Or is there any downside to this?

Swiftb0y:
Well, the most significant downside I see is that the `USBANLZ` directory can get big quite fast (its 250 MB for the 1000 Track DB I test with), which leads to significant bloat in the otherwise light repository.

Holzhaus:
Hmm, we can probably skip most of the USBANLZ files and just include a few of them.

src/pdb.rs
```diff
@@ -495,6 +495,17 @@ impl DeviceSQLString {
...
     }
 
+    fn parse_isrc_string(input: &[u8]) -> IResult<&[u8], DeviceSQLString> {
+        let (input, _) = nom::bytes::complete::tag(b"\x90")(input)?;
+        let (input, length) = nom::number::complete::le_u16(input)?;
+        let (input, _) = nom::bytes::complete::tag(b"\x00\x03")(input)?;
+        let (input, data) = nom::bytes::complete::take(length - 6)(input)?;
+        std::str::from_utf8(data).map_or_else(
+            |_| Err(nom_input_error_with_kind(input, ErrorKind::Char)),
+            |text| Ok((input, Self::ShortASCII(text.to_owned()))),
```
Holzhaus, line 505: I think we should add a new string type for this, otherwise we cannot serialize it properly when we add support for that.

Holzhaus (approved):

PR #27, opened 2022-02-16
Add parser for `*SETTING.DAT` files

Holzhaus:


Holzhaus:
Very basic because the individual settings haven't been reverse-engineered yet. But doesn't hurt to add this.

PR #28, opened 2022-02-16
pdb.rs: fix `DeviceSQLString::parse_long_utf16le` and `DeviceSQLString::parse_long_ascii`

Swiftb0y:
previous code relied on kaitai docs that theorized the presence of
two nulls at the end of these strings. The kaitai parser definition
does not check this so the theory has never been proven in practice.
Its more likely that the string just ends without any trailing null
bytes and there is likely garbage which must not be interpreted.

This theory has yet to be confirmed and it goes against the 
kaitai struct definition commonly seen as truth, so take this
PR with a grain of salt.

### DIFF
--- a/src/pdb.rs
+++ b/src/pdb.rs
@@ -512,7 +512,8 @@ impl DeviceSQLString {
         let (input, length) = nom::number::complete::le_u16(input)?;
         let (input, _) = nom::bytes::complete::tag(b"\x00")(input)?;
 
-        // sanity check
+        // The length is in bytes, UTF-16 code points are always 2 bytes wide,
+        // so a valid length must be even.
         debug_assert_eq!(length % 2, 0);
 
         let str_length = usize::from(length - 4) / 2;
--- a/src/pdb.rs
+++ b/src/pdb.rs
@@ -512,13 +512,11 @@ impl DeviceSQLString {
         let (input, length) = nom::number::complete::le_u16(input)?;
         let (input, _) = nom::bytes::complete::tag(b"\x00")(input)?;
 
-        // TODO: Ensure that length is always even
-        //assert_eq!(length % 2, 0);
-        // Don't include the trailing two nullbytes in the string.
-        let str_length = usize::from(length) / 2 - 1;
+        // sanity check
+        debug_assert_eq!(length % 2, 0);
 
+        let str_length = usize::from(length - 4) / 2;
         let (input, data) = nom::multi::count(nom::number::complete::le_u16, str_length)(input)?;
-        let (input, _) = nom::bytes::complete::tag(b"\x00\x00")(input)?;
 
         String::from_utf16(&data).map_or_else(
             |_| Err(nom_input_error_with_kind(input, ErrorKind::Char)),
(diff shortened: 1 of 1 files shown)


Holzhaus:
Both variants are pure speculation. Let's postpone the merge until we have test data to confirm one of those possibilities.

Swiftb0y:
Not really. All of the PRs I issued were fixing panics that occured in a real database. James Elliot confirmed that the two trailing nulls are also just speculation (https://deep-symmetry.zulipchat.com/#narrow/stream/275855-dysentery-.26.20crate.20digger/topic/DeviceSQLUTF16String.20trailing.20nullbytes). Assuming the database I used was not faulty to begin with, this PR makes the parser work with that database. 

Here is the relevant snippet from that DB, you can see that past the last utf16 char, you find data from a previous ASCII string. 
Screenshot from 2022-02-17 00-14-21

src/pdb.rs
```diff
@@ -512,13 +512,11 @@ impl DeviceSQLString {
         let (input, length) = nom::number::complete::le_u16(input)?;
         let (input, _) = nom::bytes::complete::tag(b"\x00")(input)?;
 
-        // TODO: Ensure that length is always even
-        //assert_eq!(length % 2, 0);
-        // Don't include the trailing two nullbytes in the string.
-        let str_length = usize::from(length) / 2 - 1;
+        // sanity check
```
Holzhaus, line 515: I think the previous comment was more descriptive tbh.

```suggestion
        //  Ensure that length is always even.
```

src/pdb.rs
```diff
@@ -512,13 +512,11 @@ impl DeviceSQLString {
...
+        debug_assert_eq!(length % 2, 0);
 
+        let str_length = usize::from(length - 4) / 2;
```
Holzhaus, line 518: Why do you subtract 4 here?

src/pdb.rs
Swiftb0y: IMO it doesnt add much value either. `length % 2` is self descriptive. A better comment would be to explain why we actually ensure this.
```suggestion
        // The length is in bytes, UTF-16 code points are always 2 bytes wide, so a valid length must be even. 
```

src/pdb.rs
Swiftb0y: Thats how its done in the reference. I don't know why these strings are off by 4 bytes, but they are. https://github.com/Deep-Symmetry/crate-digger/blob/d37097b68f186a1669c1e0358adf0cef254b9e52/src/main/kaitai/rekordbox_pdb.ksy#L945

Holzhaus: In that case, could you add a comment for this? Also, it's weird that we don't subtract 3 from the LongASCII string length then...

Swiftb0y: That might be potential bug then. Looking at the large test db, the only LongASCII occurences are `LongASCII("\u{0}")`.

Swiftb0y: Yep, pretty sure I found another error in the kaitai spec. The `LongASCII("\u{0}")` strings are actually not zero length, were just reading them one byte too early, so the string parsing stops at the null-byte. Looking at the kaitai spec, its taking that length and then reading one byte to early, and three bytes to far.

Holzhaus (approved):
Thanks. Without a test database I cannot verify these fixes, so I have to take your word for it.

Swiftb0y:
Thanks. I know not having a somewhat usable database makes testing difficult.

PR #30, opened 2022-02-17
build: Generate smoke tests for setting files

Holzhaus:


Swiftb0y:
I've been wondering about this approach to integration tests. Can you elaborate why there is this effort for a custom build process in general? Preferably document it at the top of the build.rs for future reference.

Holzhaus:
> I've been wondering about this approach to integration tests. Can you elaborate why there is this effort for a custom build process in general? Preferably document it at the top of the build.rs for future reference.

It generates the tests automatically from the contents of the data directory without having to write tests for each file individually. I have a branch where the code is made significantly easier to read using globs though.

PR #31, opened 2022-02-17
Build script improvements

Holzhaus:
This should make `build.rs` much easier to read.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3,6 +3,12 @@
 version = 3
 
 [[package]]
+name = "glob"
+version = "0.3.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "9b919933a397b79c37e33b77bb2aa3dc8eb6e165ad809e58ff75bc7db2e34574"
+
+[[package]]
 name = "memchr"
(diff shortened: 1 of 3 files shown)


Swiftb0y:
This does make the tests easier to read. However, I'm still not convinced that building a test for each database file makes sense. My preferred approach would be to just create a single test suite per file type and then parse those in that integration test. Would something like [rstest](https://docs.rs/rstest/latest/rstest/index.html) make this easier?

Holzhaus:
> However, I'm still not convinced that building a test for each database file makes sense. My preferred approach would be to just create a single test suite per file type and then parse those in that integration test.

Can you elaborate why? The point of these separate tests that it becomes very easy to see what file failed instead of "some pdb file cannot be parsed but you don't know which one". It also becomes possible to run them in parallel via `cargo test -j16` (although I admit that the tests run so fast that it makes no difference).

Swiftb0y:
Mhmm. I see. I guess the primary issue is that the tests have to be generated at build time.
You convinced me that this is a good idea. And now that you made the code legible, its maintenance burden is quite minimal.

Swiftb0y (approved):

PR #34, opened 2022-02-20
setting: Add support for parsing `DJMMYSETTING.DAT`

Holzhaus:


src/setting.rs
```diff
@@ -1128,3 +1161,86 @@ impl PadButtonBrightness {
...
+}
+
+impl ChannelFaderCurve {
+    fn parse(input: &[u8]) -> IResult<&[u8], Self> {
+        let (input, value) = nom::number::complete::u8(input)?;
+        let value = match value {
+            0x80 => Self::SteepTop,
+            0x81 => Self::Linear,
+            0x82 => Self::SteepBottom,
+            _ => Self::Unknown(value),
```
Swiftb0y, line 1186: In the long run, it would be nice if we could have some diagnostics for the `Unknown` case. This would help catch edge cases of values we don't know.

Holzhaus: Yeah, I'm not sure about it either. On the positive side, it allows deserializing and even re-serializing without information loss. On the other hand, it would probably be good to notice such issues early. Unsure what the best way to handle this is.

src/setting.rs
```diff
@@ -196,6 +208,27 @@ impl SettingData {
         }
     }
 
+    fn parse_djmmysetting(input: &[u8]) -> IResult<&[u8], Self> {
+        let (input, unknown1) = nom::bytes::complete::take(12usize)(input)?;
+        let unknown1 = unknown1.to_vec();
+        let (input, channel_fader_curve) = ChannelFaderCurve::parse(input)?;
+        let (input, crossfader_curve) = CrossfaderCurve::parse(input)?;
+        let (input, unknown2) = nom::bytes::complete::take(10usize)(input)?;
```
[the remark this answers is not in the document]
Swiftb0y: Ah yeah, my fault sorry. `ToUsize` is implemented for all unsigned numbers and rust defaults to signed I guess. IMO this is a compiler bug, but whatever.

src/setting.rs
Swiftb0y: I think we need a way to handle partial failure like this. Lots of the other unknown fields aren't arbitrary, but for example just seem to contain a static value most of the time. It would be nice to have a way of (de-)serializing them while still conveying that the contained value was differing from the expected one. This could help tremendously in helping to find databases where the usual assumptions don't apply.

Holzhaus: Well, we could add a new method (e.g. `checkForUnexpectedData() -> Result<(), Error>`) that looks for unexpected data in fields. But that is something for another PR.

src/setting.rs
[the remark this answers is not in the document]
Holzhaus: Yeah, I would expect the type to be a generic `{integer}`, not `i32`, but apparently the latter is the default integer type: https://doc.rust-lang.org/book/ch03-02-data-types.html#integer-types

PR #38, opened 2022-02-23
Add `rekordcrate_derive` crate that provides procedural macros

Holzhaus:
Replaces hundreds of lines of manual `parse` method implementations for values in `*SETTING.DAT` files with a simple `#[derive(Parse)]`. Unsure if the additional complexity of using a dedicated derive crate and the additional dependencies is worth it.

@Swiftb0y @uklotzde Feedback would be very much appreciated.

**TODO:**
- [ ] Investigate values of the `PadButtonBrightness` values. Something looks odd there.
- [ ] Check if we should use an `impl trait` instead of a direct `impl`
- [ ] Check if we should use `Deserialize` instead of `Parse` (consistent with serde)


rekordcrate/Cargo.toml
```diff
@@ -0,0 +1,19 @@
+[package]
+name = "rekordcrate"
+version = "0.1.1"
+authors = ["Jan Holthuis <jan.holthuis@ruhr-uni-bochum.de>"]
+description = "Library for parsing Pioneer Rekordbox device exports"
+readme = "README.md"
```
Holzhaus, line 6: Note to myself: I think this path is wrong now.

rekordcrate/Cargo.toml
```diff
@@ -0,0 +1,19 @@
...
+repository = "https://github.com/Holzhaus/rekordcrate"
+license = "MPL-2.0"
+keywords = ["rekordbox", "dj", "export", "database", "parser"]
+categories = ["parser-implementations"]
+edition = "2021"
+exclude = [".*"]
+
+[dependencies]
+nom = "7"
+rekordcrate_derive = { path = "../rekordcrate_derive" }
```
Holzhaus, line 16: Will this relative path work with `crates.io`? I guess it doesn't...

uklotzde: https://github.com/serde-rs/serde/blob/master/serde/Cargo.toml

Holzhaus:
Superseeded by #binrw.

Issue #39, opened 2022-02-23
Look into alternative (de-)serialization tools


Swiftb0y:
While `nom` can result in quite elegant code IMO, it only gives us deserialization. If we want to add serialization, we have to write at least an equal amount of code that does the manual serialization part. For a large part of the code base, this essentially consists of reversing match statements with little but repetitive differences spread throughout. I'm proposing to use the [binrw](https://docs.rs/binrw/latest/binrw/) crate to eliminate some of that boilerplate.

```rust
use binrw::binrw; 
#[derive(Debug, PartialEq)]
#[binrw]
#[brw(repr = u8)]
enum Test {
    A = 1,
    B = 2,
}

#[cfg(test)]
mod tests {
    use super::*;
    #[test]
    fn valid() {
        use binrw::{BinRead, io::Cursor};
        let mut cursor = Cursor::new(&[1]);
        let test = Test::read(&mut cursor).unwrap();
        assert_eq!(test, Test::A);
    }
    #[test]
    #[should_panic]
    fn invalid() {
        use binrw::{BinRead, io::Cursor};
        let mut cursor = Cursor::new(&[0]);
        Test::read(&mut cursor).unwrap();
    }
}
```
for more complex data_structures, `binrw` still has options for writing manual readers and writers. 

Moreover, writing code based on kaitai specs will also be almost as easy since the resulting macro-annotated datastructures look very similar to a kaitai spec file IMO.

Holzhaus:
Good point. I looked into `binrw` migrated the setting parser to it in #40. Please have a look.

Swiftb0y:
Great, I'll take a look, I'm also working on a PoC for the `pdb` database.

Holzhaus:
I think this can be closed now.

Swiftb0y:
Just for reference: there is also [deku](https://crates.io/crates/deku) which can work on a bit-level granularity, but I don't think its as fleshed-out as binrw + modular-bitfield.

PR #40, opened 2022-02-23
Use `binrw` instead of `nom`

Holzhaus:
Some experiments with `binrw` for #39.


src/setting.rs
```diff
@@ -17,24 +17,29 @@
 //!
 //! The exact format still has to be reverse-engineered.
 
-use crate::util::nom_input_error_with_kind;
-use nom::error::ErrorKind;
-use nom::IResult;
+use binrw::{binread, NullString};
 
 #[derive(Debug)]
+#[binread]
```
Swiftb0y, line 23: why binread instead of `binrw`?

src/setting.rs
```diff
@@ -17,24 +17,29 @@
...
+#[br(little)]
 /// Represents a setting file.
 pub struct Setting {
     /// Size of the string data field (should be always 96).
+    #[br(assert(len_stringdata == 0x60))]
     pub len_stringdata: u32,
     /// Name of the company ("PIONEER").
-    pub company: String,
+    #[br(pad_size_to = 0x20)]
+    pub company: NullString,
```
Swiftb0y, line 32:
```suggestion
    #[br(pad_size_to = 0x20)]
    #[brw(magic = b"PIONEER\x00")]
```
I'm actually not sure how `magic` works or is intended to be used, but I think we get writing for free.

src/setting.rs
```diff
@@ -201,1380 +169,781 @@ pub enum SettingData {
...
 /// Rekordbox preferences.
 #[derive(Debug)]
+#[binread]
 pub enum PlayMode {
     /// Named "CONTINUE / ON" in the Rekordbox preferences.
+    #[br(magic = 0x80u8)]
     Continue,
     /// Named "SINGLE / OFF" in the Rekordbox preferences.
+    #[br(magic = 0x81u8)]
     Single,
```
Swiftb0y, line 197: I think the preferred idiom is using a unit-like struct
```suggestion
#[derive(Debug)]
#[binread]
#[brw(repr(u8))]
pub enum PlayMode {
    /// Named "CONTINUE / ON" in the Rekordbox preferences.
    Continue = 0x80,
    /// Named "SINGLE / OFF" in the Rekordbox preferences.
    Single,
```

Swiftb0y (changes requested):

src/setting.rs
Holzhaus: `magic` only works with fixed values. But This is sometimes `PIONEER`, sometimes `PIONEERDJ`.

Swiftb0y: Thanks, this should be documented then I guess.

Holzhaus: I've seen these values:

```bash
$ for file in data/*/PIONEER/*.DAT; do cargo run --bin rekordcrate-setting -- "$file"; done | grep String
    Finished dev [unoptimized + debuginfo] target(s) in 0.01s
     Running `target/debug/rekordcrate-setting data/demo_tracks/PIONEER/DEVSETTING.DAT`
    company: NullString("PIONEER DJ"),
    software: NullString("rekordbox"),
    version: NullString("6.6.1"),
    Finished dev [unoptimized + debuginfo] target(s) in 0.02s
     Running `target/debug/rekordcrate-setting data/demo_tracks/PIONEER/DJMMYSETTING.DAT`
    company: NullString("PioneerDJ"),
    software: NullString("rekordbox"),
    version: NullString("1.000"),
    Finished dev [unoptimized + debuginfo] target(s) in 0.02s
     Running `target/debug/rekordcrate-setting data/demo_tracks/PIONEER/MYSETTING2.DAT`
    company: NullString("PIONEER"),
    software: NullString("rekordbox"),
    version: NullString("0.001"),
    Finished dev [unoptimized + debuginfo] target(s) in 0.02s
     Running `target/debug/rekordcrate-setting data/demo_tracks/PIONEER/MYSETTING.DAT`
    company: NullString("PIONEER"),
    software: NullString("rekordbox"),
    version: NullString("0.001"),
    Finished dev [unoptimized + debuginfo] target(s) in 0.02s
     Running `target/debug/rekordcrate-setting data/empty/PIONEER/DEVSETTING.DAT`
    company: NullString("PIONEER DJ"),
    software: NullString("rekordbox"),
    version: NullString("6.6.1"),
    Finished dev [unoptimized + debuginfo] target(s) in 0.02s
     Running `target/debug/rekordcrate-setting data/empty/PIONEER/DJMMYSETTING.DAT`
    company: NullString("PioneerDJ"),
    software: NullString("rekordbox"),
    version: NullString("1.000"),
    Finished dev [unoptimized + debuginfo] target(s) in 0.02s
     Running `target/debug/rekordcrate-setting data/empty/PIONEER/MYSETTING2.DAT`
    company: NullString("PIONEER"),
    software: NullString("rekordbox"),
    version: NullString("0.001"),
    Finished dev [unoptimized + debuginfo] target(s) in 0.02s
     Running `target/debug/rekordcrate-setting data/empty/PIONEER/MYSETTING.DAT`
    company: NullString("PIONEER"),
    software: NullString("rekordbox"),
    version: NullString("0.001"),
```

src/setting.rs
Holzhaus: https://github.com/Holzhaus/rekordcrate/pull/40/commits/d745bb71165cd4ee270332b269b94101a1e14b7c

src/setting.rs
Holzhaus: I wanted to archieve feature partity first. It's easy to change using search+replace, but I wasn't sure how much extra work was needed for serialization. Turns out that it's almost for free: https://github.com/Holzhaus/rekordcrate/pull/40/commits/eec1aace3a9ff18966c0ce7af847d4d1d69d5edd

src/setting.rs
```diff
@@ -17,24 +17,38 @@
...
+    pub software: NullString,
     /// Some kind of version number.
-    pub version: String,
+    #[brw(pad_size_to = 0x20)]
+    pub version: NullString,
     /// Size of the `data` data in bytes.
     pub len_data: u32,
-    /// Unknown field.
+    /// The actual settings data.
+    #[br(args(len_data))]
```
Swiftb0y, line 51:
```suggestion
    #[brw(args(len_data))]
```

src/setting.rs
Holzhaus: https://github.com/Holzhaus/rekordcrate/pull/40/commits/d349f779c024897ed7eec8d0526bcd316ed090b1

src/setting.rs
```diff
@@ -44,65 +58,24 @@ pub struct Setting {
     /// details.
     pub checksum: u16,
     /// Unknown field (apparently always `0000`).
+    #[br(assert(unknown == 0))]
```
Swiftb0y, line 61:
```suggestion
    #[brw(assert(unknown == 0))]
```

src/setting.rs
```diff
@@ -201,636 +182,271 @@ pub enum SettingData {
...
-        let (input, value) = nom::number::complete::u8(input)?;
-        let value = match value {
-            0x80 => Self::Off,
-            0x81 => Self::On,
-            0x82 => Self::RekordboxSetting,
-            _ => Self::Unknown(value),
-        };
-        Ok((input, value))
-    }
+    On = 0x81,
```
Swiftb0y, line 263: consider reordering

src/setting.rs
```diff
@@ -201,636 +182,271 @@ pub enum SettingData {
...
 /// Rekordbox preferences.
-#[derive(Debug)]
+#[derive(Debug, PartialEq)]
+#[binrw]
+#[brw(repr = u8)]
 pub enum PlayMode {
     /// Named "CONTINUE / ON" in the Rekordbox preferences.
-    Continue,
+    Continue = 0x80,
     /// Named "SINGLE / OFF" in the Rekordbox preferences.
```
Swiftb0y, line 208:
```rust
#[brw(repr = i8)]
pub enum PlayMode {
    /// Named "CONTINUE / ON" in the Rekordbox preferences.
    Continue = i8::MIN,
```
IMO also a viable option of expressing the same thing. I'm not sure why the values start at `0x80`. Could either be that the enums start at the lowest value and are signed (for whatever reason), or they just carry some extra info in their highest bit. Idk, feel free to reject this suggestion, I just wanted to throw the thought out there.

Swiftb0y (changes requested):

src/setting.rs
Holzhaus: This prevents serializing other values though. Do we really want that?

src/setting.rs
Holzhaus: It would be interesting if we had some files from older rekordbox versions. I have a hunch that the highest bit indicates if a field is in use.

src/setting.rs
Holzhaus: At the moment the order is the same as in the rekordbox preferences. We'd lose that info if we reorder.

src/setting.rs
Holzhaus: Nope, that doesn't make sense. The length is needed for deserialization (to decide which data type is used), but when writing the byte size is determined by the data type. It doesn't make sense to pass the length data.

build.rs
```diff
@@ -63,6 +63,8 @@ use rekordcrate::anlz::ANLZ;
         out_dir.join("tests_setting.rs"),
         "data/*/PIONEER/*.DAT",
         r#"// THIS FILE IS AUTOGENERATED - DO NOT EDIT!
+use std::io::Cursor;
+use binrw::{{BinRead, BinWrite}};
```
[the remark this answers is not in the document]
Holzhaus: https://github.com/Holzhaus/rekordcrate/pull/40/commits/3f2e76273a5403c535fea4829334bc4b3a57292e

src/setting.rs
Swiftb0y: Well, do we have a reason to serialize other values right now? IMO we can still change that and lifting that restriction is not a breaking change.

src/setting.rs
Swiftb0y: you're right, it doesn't make much sense. Though I'm unsure whether it actually serializes it correctly in the current state, I'll have to check.

src/setting.rs
Swiftb0y: Mhmm. IMO ordering it `ON`, `OFF`, `RekordboxSetting` makes more sense semantically though. I don't think there is much value in retaining the same order as in the rekordbox preferences. It doesn't make a difference semantically and I also don't see a reason for keeping that ordering for documentation purposes. After all, the (imo) weird ordering of these option in the UI, could just be a UI bug.

src/setting.rs
Swiftb0y: Pioneer still hosts older rekordbox versions: https://rekordbox.com/en/support/faq/about-rekordbox-5/ (see the "I want to use previous rekordbox ver. 5." Question).

Swiftb0y: can you elaborate what you mean with "if a field is in use"? In use by what?

Holzhaus: > can you elaborate what you mean with "if a field is in use"? In use by what?

A bunch of fields/bytes are just 0x00. But every actual value is > 0x80. I guess that the 0x00 bytes are reserved for future use. 

If a new CDJ comes out, Pioneer might decide to add new config options. If you make a fresh export using the latest Rekordbox version, the generated DAT will contain the selected value for that new config option. 

Now let's say someone is still using RB5, then thr generated config files will not contain that config option and the field will just be 0x00 ("not set"). The CDJ can detect that and use a sensible default instead.

Swiftb0y: Sounds plausible, but on the other hand, I'd be very surprised if Pioneer cared about any sort of backwards compatibility with RB5. Its more plausible that they want to maintain compatibility with old CDJs, but your theory does not work out for preserving compatibility in that direction.

src/setting.rs
Holzhaus: > Well, do we have a reason to serialize other values right now?

In theory, a user could fill in the values by hand and then serialize it.

> IMO we can still change that and lifting that restriction is not a breaking change.

Sure. Probably makes sense to add it.

Unfortunately I can't get it to work. Sometimes `unknown` is an `u16`, sometimes it's apparently an `&u16`?!

```rust
error[E0277]: can't compare `&u16` with `{integer}`
  --> src/setting.rs:61:26
   |
61 |     #[brw(assert(unknown == 0))]
   |                          ^^ no implementation for `&u16 == {integer}`
   |
```

```rust
error[E0277]: can't compare `&u16` with `u16`
  --> src/setting.rs:61:26
   |
61 |     #[brw(assert(unknown == 0u16))]
   |                          ^^ no implementation for `&u16 == u16`
   |
```

```rust
error[E0308]: mismatched types
  --> src/setting.rs:61:29
   |
61 |     #[brw(assert(unknown == &0u16))]
   |                             ^^^^^ expected `u16`, found `&u16`
   |
```

```rust
error[E0614]: type `u16` cannot be dereferenced
  --> src/setting.rs:61:18
   |
61 |     #[brw(assert(*unknown == 0u16))]
   |                  ^^^^^^^^
```

If you want to have a look, feel free to do so :D

src/setting.rs
Holzhaus: I can't really judge if this is an UI bug or not, but this is how it looks in RB6:

image

I don't think it hurts readability that much, let's just keep it for now.

src/setting.rs
Holzhaus: It does, I added round-trip tests. But it's possible to input "wrong" values. Ideally, this field would be temp and automatically calculated.

src/setting.rs
Holzhaus: For the other direction this is not necessary. When the CDJ is released, it supports a specific set of options. If Pioneer adds more options, the CDJ can just ignore these bytes.

Holzhaus: > I'd be very surprised if Pioneer cared about any sort of backwards compatibility with RB5.

It would kind of suck if you plug your prepared USB drive into the CDJs at a venue and it doesn't work because the drive was created using an old RB version.

>  Its more plausible that they want to maintain compatibility with old CDJs, but your theory does not work out for preserving compatibility in that direction.

For the other direction this is not necessary. The firmware of the CDJ knows the set of options it supports, and can safely ignore all other bytes in the file, even if more recent rekordbox versions start using them for new options.

Holzhaus:
Let's merge, the other issues are just minor things. Merging now makes it possible to open PRs yourself and saving yourself a detour over me :D

src/setting.rs
Swiftb0y: Mhmm, I think thats a bug. I feel like `brw` just gets expanded into a `br` and `bw` macro which have different types for their asserts then or something.

src/setting.rs
Swiftb0y: > It would kind of suck if you plug your prepared USB drive into the CDJs at a venue and it doesn't work because the drive was created using an old RB version.

I fully agree, but Pioneer is a profit-driven company that sells not only their hardware products, but also their rekordbox software, so I'd be pretty surprised if they actually have the decency to support RB5 when they could just force their consumers to subscribe to RB6. After all, they essentially have a monopoly in the professional DJ space.

Swiftb0y: > For the other direction this is not necessary. The firmware of the CDJ knows the set of options it supports, and can safely ignore all other bytes in the file, even if more recent rekordbox versions start using them for new options.

That is only possible if the binary file is structured appropriately. I have not taken my time to understand it to the point where I can evaluate that, but I guess you can.

Swiftb0y:
> Let's merge, the other issues are just minor things. Merging now makes it possible to open PRs yourself and saving yourself a detour over me :D

Sure, I'll issue a PR for the rest of my nitpicks.

PR #43, opened 2022-02-24
setting: Add tests for individual setting values

Holzhaus:


tests/test_djmmysetting.rs
```diff
@@ -0,0 +1,167 @@
...
+        result.expect("failed to match data section")
+    }};
+}
+
+#[test]
+fn read_djmsetting_beatfxquantize_off() {
+    let data = read_djmmysetting!("../data/djmmysetting/beatfxquantize_off/DJMMYSETTING.DAT");
+    assert_eq!(data.beat_fx_quantize, BeatFXQuantize::Off);
+}
+
```
Swiftb0y, line 23: I'm questioning this approach slightly. These are essentially unit tests but running on an entire settings file.

tests/test_mysetting.rs
```diff
@@ -0,0 +1,207 @@
...
+    ($path:literal) => {{
+        let data = include_bytes!($path);
+        println!("Setting file: {}", $path);
+        let mut reader = Cursor::new(data);
+        let setting = Setting::read(&mut reader).expect("failed to parse setting file");
+        let result = match setting.data {
+            SettingData::MySetting(x) => Some(x),
+            _ => None,
+        };
+        result.expect("failed to match data section")
```
Swiftb0y, line 14:
```suggestion
        assert!(matches!(setting.data, SettingData::MySetting(_)), "failed to match data section") 
```

Holzhaus: This asserts that `setting.data` is of the expected type, but doesn't actually destructure. The macro still returns a variable of type `SettingData`, whereas the original version returns a `MySetting` struct.

tests/test_djmmysetting.rs
Holzhaus: There are two reasons for this:
1. It's very easy to create these kinds of tests, because this is data that comes directly from rekordbox, after resetting all settings to defaults and only changing that single option. In theory, you should even be able to verify the correctness of the test data by copying that file back to rekordbox or onto a USB drive and plug it into a device to check that the settings work.
2. The settings basically map a single byte to enum variant. If the test just parsed the data without the full file, we'd more or less just duplicate that mapping into a test.

tests/test_mysetting.rs
Swiftb0y: good point, I didn't consider the function in its outer context.

tests/test_djmmysetting.rs
Swiftb0y: I still don't really understand it. What exactly are these tests supposed to test? I assume that the entire parser works correctly? If so, why do we test so many files and not just a single snapshot?

Holzhaus: > I assume that the entire parser works correctly?

No, that's just a side effect. It's supposed to check if the parsed value matches the expected value, to make sure that the mapping between enum values and bytes are correct.

> If so, why do we test so many files and not just a single snapshot?

These files are basically a byproduct of the reverse engineering. And a single snapshot wouldn't suffice to check multiple combinations anyway (a lot of fields have more than 2 possible values).

Holzhaus:
Btw, these tests help me find the issues from #42.

tests/test_djmmysetting.rs
Swiftb0y: > It's supposed to check if the parsed value matches the expected value, to make sure that the mapping between enum values and bytes are correct.

Mhmm, then I don't see much use for them to be honest. I think those mappings are set in stone and won't change anytime soon so I don't see a great amount of value in checking for them. 

But I digress, the work is already done and I don't think these tests will require a great deal of maintenance, so we might as well keep them.

Holzhaus: Well. it certainly helped to find the bugs I fixed in https://github.com/Holzhaus/rekordcrate/pull/42. I think the value is mostly verifying that nothing breaks in the future if we touch these files again. It doesn't hurt IMHO.

PR #44, opened 2022-02-24
setting: Implement `Default` trait for data structs

Holzhaus:
Based on #43.

### DIFF
--- a/src/setting.rs
+++ b/src/setting.rs
@@ -16,6 +16,11 @@
 //! curve, etc.
 //!
 //! The exact format still has to be reverse-engineered.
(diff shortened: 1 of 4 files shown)


Swiftb0y:
Have you considered using another macro crate for reducing the boilerplate? https://crates.io/crates/smart-default

Swiftb0y (approved):
Thought about this some more and I share the opinions of other library authors that the increased compilation-time from pulling in yet another proc-macro does not make sense just to save a couple lines.

Swiftb0y (approved):

PR #47, opened 2022-02-24
anlz: Use `binrw` instead of `nom` to parse `ANLZ*.DAT` files

### LINKED ISSUE #46: Port `anlz.rs` from nom to binrw

Holzhaus:
Closes #46.

### DIFF
--- a/src/anlz.rs
+++ b/src/anlz.rs
@@ -531,188 +531,264 @@ pub struct Phrase {
 /// Section content which differs depending on the section type.
 #[derive(Debug, PartialEq)]
 #[binread]
-#[br(big, import(header: Header))]
+#[br(import(header: Header))]
 pub enum Content {
     /// All beats in the track.
     #[br(pre_assert(header.kind == ContentKind::BeatGrid))]
-    BeatGrid {
-        /// Unknown field.
-        unknown1: u32,
-        /// Unknown field.
-        ///
-        /// According to [@flesniak](https://github.com/flesniak), this is always `00800000`.
-        unknown2: u32,
-        /// Number of beats in this beatgrid.
-        len_beats: u32,
-        /// Beats in this beatgrid.
-        #[br(count = len_beats)]
-        beats: Vec<Beat>,
-    },
+    BeatGrid(BeatGrid),
     /// List of cue points or loops (either hot cues or memory cues).
     #[br(pre_assert(header.kind == ContentKind::CueList))]
-    CueList {
-        /// The types of cues (memory or hot) that this list contains.
-        list_type: CueListType,
-        /// Unknown field
-        unknown: u16,
-        /// Number of cues.
-        len_cues: u16,
-        /// Unknown field.
-        memory_count: u32,
-        /// Cues
-        #[br(count = len_cues)]
-        cues: Vec<Cue>,
-    },
+    CueList(CueList),
     /// List of cue points or loops (either hot cues or memory cues, extended version).
     ///
     /// Variation of the original `CueList` that also adds support for more metadata such as
     /// comments and colors. Introduces with the Nexus 2 series players.
     #[br(pre_assert(header.kind == ContentKind::ExtendedCueList))]
-    ExtendedCueList {
-        /// The types of cues (memory or hot) that this list contains.
-        list_type: CueListType,
-        /// Number of cues.
-        len_cues: u16,
-        /// Unknown field
-        #[br(assert(unknown == 0))]
-        unknown: u16,
-        /// Cues
-        #[br(count = len_cues)]
-        cues: Vec<ExtendedCue>,
-    },
+    ExtendedCueList(ExtendedCueList),
     /// Path of the audio file that this analysis belongs to.
     #[br(pre_assert(header.kind == ContentKind::Path))]
-    Path {
-        /// Length of the path field in bytes.
-        len_path: u32,
-        #[br(assert(len_path == header.content_size()), count = len_path)]
-        /// Path of the audio file.
-        path: Vec<u8>,
-    },
+    Path(#[br(args(header.clone()))] Path),
     /// Seek information for variable bitrate files (probably).
     #[br(pre_assert(header.kind == ContentKind::VBR))]
-    VBR {
-        /// Unknown field.
-        unknown1: u32,
-        /// Unknown data.
-        #[br(count = header.content_size())]
-        unknown2: Vec<u8>,
-    },
+    VBR(#[br(args(header.clone()))] VBR),
     /// Fixed-width monochrome preview of the track waveform.
     #[br(pre_assert(header.kind == ContentKind::WaveformPreview))]
-    WaveformPreview {
-        /// Unknown field.
-        len_preview: u32,
-        /// Unknown field (apparently always `0x00100000`)
-        unknown: u32,
-        /// Waveform preview column data.
-        #[br(count = header.content_size())]
-        data: Vec<WaveformPreviewColumn>,
-    },
+    WaveformPreview(#[br(args(header.clone()))] WaveformPreview),
     /// Smaller version of the fixed-width monochrome preview of the track waveform.
     #[br(pre_assert(header.kind == ContentKind::TinyWaveformPreview))]
-    TinyWaveformPreview {
-        /// Unknown field.
-        len_preview: u32,
-        /// Unknown field (apparently always `0x00100000`)
-        unknown: u32,
-        /// Waveform preview column data.
-        #[br(count = header.content_size())]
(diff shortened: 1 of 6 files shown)


src/bin/rekordcrate-anlz.rs
```diff
@@ -6,14 +6,15 @@
...
-    let (input, anlz) = ANLZ::parse(&data).expect("failed to parse header");
+    let mut reader = std::fs::File::open(&path).expect("failed to open file");
+    let anlz = ANLZ::read(&mut reader).expect("failed to parse setting file");
     println!("File Header: {:#?}", anlz);
-    for (i, section) in anlz.sections(input).enumerate() {
-        println!("#{}: {:#?}", i, section);
-    }
+    //for (i, section) in anlz.sections(input).enumerate() {
+    //    println!("#{}: {:#?}", i, section);
+    //}
```
Swiftb0y, line 19: Why did you comment this out?

src/util.rs
```diff
@@ -18,26 +19,36 @@ pub fn nom_input_error_with_kind(input: &[u8], kind: ErrorKind) -> Err<nom::erro
     Err::Error(nom::error::Error::from_error_kind(input, kind))
 }
 
-#[derive(Debug)]
 /// Indexed Color identifiers used for memory cues and tracks.
+#[derive(Debug, PartialEq)]
+#[binread]
```
Swiftb0y, line 24: this could be `binrw`

src/bin/rekordcrate-anlz.rs
Holzhaus: Ah, it's obsolete because all sections are parsed upfront now.

Holzhaus: I'll remove it.

src/util.rs
Holzhaus: Yeah, as before I'd like to get read support finished first, before dealing with serialization. If no changes are needed, `sed -i 's/[binread]/[binrw]/g' -e 's/[br(/[brw(/g' anlz.rs` will probably suffice.

src/anlz.rs
```diff
@@ -359,8 +261,11 @@ pub struct ExtendedCue {
     pub loop_numerator: u16,
     /// Represents the loop size denominator (if this is a quantized loop).
     pub loop_denominator: u16,
-    /// And UTF-16BE encoded string, followed by a trailing  `0x0000`.
-    pub comment: String,
+    /// Length of the comment string in bytes.
+    pub len_comment: u32,
+    /// An UTF-16BE encoded string, followed by a trailing  `0x0000`.
+    #[br(count = len_comment)]
+    pub comment: Vec<u8>,
```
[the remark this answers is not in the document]
Holzhaus: Hmm, I'm unsure how to handle it properly. We need to make sure that we exactly parse that many bytes.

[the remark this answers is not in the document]
Swiftb0y: Do you mean we can't rely on the fact that the string is null-terminated or that we need to make sure that the parsed string was only that many bytes long?
For the latter cased, an `assert` is likely sufficient, for the former case, I used something like this for the `pdb`:
```rust
UCS2(
  #[br(count = len / 2)]
  #[br(try_map = |x: Vec<u16>| String::from_utf16(&x))]
  #[bw(map = |x: &String| x.encode_utf16().collect::<Vec<u16>>())]
  String
)
```
Note: unfortunately using `&String` is required otherwise the macro trips up somehow.

We should probably factor this out into a separate type similar to `NullString` and `NullWideString` and then maybe file a PR to binrw so they get included. This case seems common enough.

src/bin/rekordcrate-anlz.rs
Swiftb0y: So they are parsed as part of the FileHeader?

src/util.rs
Swiftb0y: Thats what I thought about the not as trivial types, but making this type serializeable is just a matter of changing this one line, which is why I proposed it.

src/bin/rekordcrate-anlz.rs
Holzhaus: Not really. Basically the file is structured in sections consisting of Header and Content. There is exactly 1 outer section, and the content part contains the other sections. I introduced this separation and only parsed the header of the outer section because I thought this would be more memory efficient. Considering that these files are 5 - 100KiB for the demo tracks, I don't think thats a big issue though.

src/anlz.rs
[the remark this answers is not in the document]
Holzhaus: > Do you mean we can't rely on the fact that the string is null-terminated or that we need to make sure that the parsed string was only that many bytes long?
> For the latter cased, an assert is likely sufficient, for the former case, I used something like this for the pdb: [...]

I used a `NullWideString` for now. In theory, there should always be a nullbyte. The issue is that I always learned "do not trust user data". Depending on a nullbyte seems wrong because a maliciously crafted file may omit this and lead to freeze/OOM issues or something like that. And that would be unnecessary because we *know* beforehand how many bytes should be read.

IMHO the cleanest way would be to introduce a new attribute that limits the maximum number of bytes that can be read. That would be reusable even for other kinds of data structures.

```rust
pub struct Foo {
    len_text: u32,
    #[br(limit = len_text)]
    text: NullString,
}

pub struct Bar {
    #[br(assert(len_text % 2 == 0))]
    len_text: u32,
    #[br(limit = len_text)]
    text: NullWideString,
}

File: src/util.rs (hunk 2 above)
Comment (Holzhaus) [reply src/util.rs]: https://github.com/Holzhaus/rekordcrate/pull/47/commits/f0d4ad0343e0ef0c57f36ba6f20f884d2403500e

File: src/anlz.rs [hunk 4]
```diff
@@ -24,87 +24,83 @@
...
+    #[brw(magic = b"PWV5")]
     WaveformColorDetail,
     /// Describes the structure of a sond (Intro, Chrous, Verse, etc.).
     ///
     /// Used in `.EXT` files.
+    #[brw(magic = b"PSSI")]
     SongStructure,
     /// Unknown Kind.
     Unknown([u8; 4]),
 }
```
Swiftb0y, line 98: Could you add a short explanation for each Unknown why that couldn't be removed?

src/anlz.rs
```diff
@@ -775,522 +531,315 @@ pub struct Phrase {
...
+    unknown1: u32,
+    /// Unknown field.
+    ///
+    /// According to [@flesniak](https://github.com/flesniak), this is always `00800000`.
+    unknown2: u32,
+    /// Number of beats in this beatgrid.
+    len_beats: u32,
+    /// Beats in this beatgrid.
+    #[br(count = len_beats)]
+    pub beats: Vec<Beat>,
```
Swiftb0y, line 602: note that exposing the `Vec` publicly mutably means that the `beats.len() == len_beats` invariant can be violated. `binrw` currently does not have a way to fix these up easily. see the issue where this is discussed in the context of `binrw::FilePtr`: https://github.com/jam1garner/binrw/issues/4

Holzhaus: > note that exposing the `Vec` publicly mutably means that the `beats.len() == len_beats` invariant can be violated.

Isn't that the case anyway even if the field was not `pub`? I could still crate an instance of the struct with mismatching len_beats and beats values by hand. The fix would be to make `len_beats` a temp field that isn't actually part of the struct and is only parsed for reading and automatically calculated from `beats` when writing.

src/anlz.rs
Holzhaus: https://github.com/Holzhaus/rekordcrate/pull/47/commits/04b987f73ce5c4cb3bccd63cf38dd67806cf9c64

Holzhaus: Without this, we wouldn't be able to parse `*.2EX` files. See #9.

src/anlz.rs
Swiftb0y: Not sure, as mentioned in the FilePtr issue, a second pass could probably help with most of these issues. Another one could be to pass a reference to the field in the types args, but that is not supported either currently. Though both are on the v0.9.0 roadmap https://github.com/users/jam1garner/projects/1 

> Isn't that the case anyway even if the field was not pub? I could still crate an instance of the struct with mismatching len_beats and beats values by hand.

well if the type is public then yes, for this reason, I opted for an encapsulation approach in the DeviceSQLString implementation where the string is constructed correctly once and then should never be manipulated again, at least thats the plan, there are still some kinks to work out.

Holzhaus: Should be fixed with https://github.com/Holzhaus/rekordcrate/pull/51.

PR #48, opened 2022-02-25
feat(pdb): Add binrw-based DeviceSQLString impl

Swiftb0y:
I still have to reiterate a bit on the design, especially since I already found flaws while creating unit tests (also notice all the TODOs)

### DIFF
--- a/src/pdb/string.rs
+++ b/src/pdb/string.rs
@@ -263,6 +263,24 @@ mod test {
     }
 
     #[test]
+    fn too_long_string() {
+        // construct super long string containing just "AAAAAAA"...
+        const TOO_LARGE_STR_SIZE: usize = (u16::MAX as usize) + 1;
+        const INIT_CHAR: char = 'A';
+        const _: () = assert!(INIT_CHAR.is_ascii());
+        const HUMONGOUS_ARRAY: [u8; TOO_LARGE_STR_SIZE] = [INIT_CHAR as u8; TOO_LARGE_STR_SIZE];
+
+        // Since we already know that the string only contains ascii at compile time,
+        // we could probably skip the validation, but that requires unsafe code which I'd consider overkill.
+        let humongous_string = String::from_utf8(HUMONGOUS_ARRAY.to_vec()).unwrap();
+
+        assert_eq!(
+            DeviceSQLString::new(humongous_string).unwrap_err(),
+            StringError::TooLong
+        );
+    }
+
+    #[test]
     fn isrc_edge_case() -> Result<(), StringError> {
         let serialized = [
             0x90, 0x12, 0x00, 0x00, 0x03, 0x47, 0x42, 0x41, 0x59, 0x45, 0x36, 0x37, 0x30, 0x30,
@@ -273,6 +291,12 @@ mod test {
             DeviceSQLString::new_isrc("GBAYE6700149".to_string())?,
         );
         test_roundtrip(&[0x3], DeviceSQLString::new_isrc("".to_string())?);
+
+        assert_eq!(
+            DeviceSQLString::new_isrc("non-conforming garbage".to_string()).unwrap_err(),
+            StringError::InvalidISRC
+        );
+
         Ok(())
     }
 }
--- a/src/pdb/string.rs
+++ b/src/pdb/string.rs
@@ -1,0 +1,261 @@
+// Copyright (c) 2022 Nikolaus Einhauser
+//
+// This Source Code Form is subject to the terms of the Mozilla Public License, v. 2.0. If a copy
+// of the MPL was not distributed with this file, You can obtain one at
+// http://mozilla.org/MPL/2.0/.
+//
+// SPDX-License-Identifier: MPL-2.0
+
+//! `binrw`-based implementation for DeviceSQLStrings capable of parsing and
+//! serializing [`DeviceSQLString`]s
+//!
+//! See <https://djl-analysis.deepsymmetry.org/rekordbox-export-analysis/exports.html#devicesql-strings>
+
+use binrw::{binrw, NullString};
+
+const MAX_SHORTSTR_SIZE: usize = ((u8::MAX >> 1) - 1) as usize;
+
+/// Encapsulates the intrinsics of the format used by DeviceSQL strings
+///
+/// Once A [`DeviceSQLString`] has been constructed, there is no way to change it.
+///
+/// ```
+/// # pub fn main() -> binrw::BinResult<()> {
+/// use rekordcrate::pdb::string::DeviceSQLString;
+/// use binrw::{BinWrite, BinRead};
+/// let string = DeviceSQLString::new("foo".to_owned());
+/// let binary = vec![0x9, 0x66, 0x6F, 0x6F];
+///
+/// let mut writer = binrw::io::Cursor::new(vec![]);
+/// string.write_to(&mut writer)?;
+/// assert_eq!(&binary, writer.get_ref());
+///
+/// let mut reader = binrw::io::Cursor::new(binary);
+/// let parsed = DeviceSQLString::read(&mut reader)?;
+/// assert_eq!(parsed, string);
+/// # Ok(())
+/// # }
+/// ```
+#[derive(Debug, PartialEq)]
+#[binrw]
+pub struct DeviceSQLString(DeviceSQLStringImpl);
+impl DeviceSQLString {
+    /// Initializes a [`DeviceSQLString`] from a plain Rust [`std::string::String`]
+    #[must_use]
+    pub fn new(string: String) -> Self {
+        let len = string.len();
+        let only_ascii = string.is_ascii();
+        if only_ascii && len <= MAX_SHORTSTR_SIZE {
+            Self(DeviceSQLStringImpl::ShortASCII {
+                content: string.into_bytes(),
+            })
+        } else if len <= (i16::MAX as usize) {
+            if only_ascii {
+                Self(DeviceSQLStringImpl::Long {
+                    content: LongBody::Ascii(string.into_bytes()),
+                })
+            } else {
+                Self(DeviceSQLStringImpl::Long {
+                    // note: The DeviceSQL database may only support UCS-2 so
+                    // we might need to do some additional filtering here
+                    content: LongBody::Ucs2le(string.encode_utf16().collect()),
+                })
+            }
+        } else {
+            todo!()
+        }
+    }
+
+    /// Creates a [`DeviceSQLString`] containing an ISRC instead of an expected string
+    ///
+    /// The DeviceSQL Database as used by rekordbox has a strange oddity
+    /// where it serializes strings containing a tracks ISRC (International
+    /// Standard Recording Code) in an unexpected format, if this is desired,
+    /// use this constructor function instead of [`DeviceSQLString::new`].
+    #[must_use]
+    pub fn new_isrc(string: String) -> Self {
+        if string.is_empty() {
+            return Self::empty();
+        }
+        let len = string.len();
+        // basic validation taken from
+        // https://isrc.ifpi.org/downloads/ISRC_Bulletin-2015-01.pdf
+        debug_assert!(string.is_ascii());
+        debug_assert_eq!(len, 12);
+        Self(DeviceSQLStringImpl::Long {
+            content: LongBody::Isrc(NullString::from_string(string)),
+        })
+    }
+
+    /// Extract the Rust string from the DeviceSQLString.
+    ///
+    /// Consumes itself in the process.
+    #[must_use]
+    pub fn into_string(self) -> String {
+        match self.0 {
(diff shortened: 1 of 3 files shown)


src/string.rs
```diff
@@ -0,0 +1,161 @@
...
+
+#[derive(Debug, PartialEq)]
+#[binrw]
+struct DeviceSQLShortASCIIString {
+    #[br(assert(header & 0b1 == 1))]
+    header: u8,
+    #[br(count = (header >> 1) - 1)]
+    #[br(try_map = String::from_utf8)]
+    #[bw(map = |x: &String| x.as_bytes())]
+    content: String,
```
Holzhaus, line 63: `NullString` uses a `Vec<u8>` internally and then provides `into_string`/`into_string_lossless` methods. Any particular reason why you chose a different approach? https://docs.rs/binrw/latest/src/binrw/strings.rs.html#57-60

Swiftb0y: Yeah, I only found out about that after the fact, I'll have to look into the API ergonomics, I might consider adopting it.

Swiftb0y:
This is ready for review now

src/pdb/string.rs
```diff
@@ -0,0 +1,275 @@
...
+
+    // ISRC strings are a bug/flaw in pioneers implementation, this is technically
+    // a semi-dirty workaround
+    #[br(pre_assert(flags == 0x90))]
+    Isrc(#[brw(magic = 0x3u8)] binrw::NullString),
+    #[br(pre_assert(flags == 0x40))]
+    Ascii(#[br(count = len)] Vec<u8>),
+    #[br(pre_assert(flags == 0x90))]
+    #[br(pre_assert(len % 2 == 0))]
+    Ucs2le(#[br(count = len / 2)] Vec<u16>),
```
Holzhaus, line 182: These enum members should be uppercase IMHO. 

```suggestion
    #[br(pre_assert(flags == 0x90))]
    ISRC(#[brw(magic = 0x3u8)] binrw::NullString),
    #[br(pre_assert(flags == 0x40))]
    ASCII(#[br(count = len)] Vec<u8>),
    #[br(pre_assert(flags == 0x90))]
    #[br(pre_assert(len % 2 == 0))]
    UCS2LE(#[br(count = len / 2)] Vec<u16>),
```

src/pdb/string.rs
```diff
@@ -0,0 +1,275 @@
...
+}
+
+impl LongBody {
+    pub fn byte_count(&self) -> Result<u16, StringError> {
+        match self {
+            // isrc offset is compensating for trailing nullbyte + 0x3 magic byte
```
Holzhaus, line 188:
```suggestion
            // ISRC offset is compensating for trailing nullbyte + 0x3 magic byte. 
```

Holzhaus (commented):
Some minor nits.

src/util.rs
```diff
@@ -84,3 +84,31 @@ impl From<u16> for ColorIndex {
...
 }
+
+#[cfg(test)]
+pub(in crate) mod testing {
+    use binrw::prelude::*;
+    pub fn test_roundtrip<T>(bin: &[u8], obj: T)
+    where
+        <T as binrw::BinRead>::Args: Default,
+        <T as binrw::BinWrite>::Args: Default,
+        T: BinRead + BinWrite + PartialEq + core::fmt::Debug,
```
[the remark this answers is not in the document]
Swiftb0y: It is required for the asserts. In case an assert fails, it prints the debug representation of the passed object.

src/pdb/string.rs
Swiftb0y: `clippy::all` implies `clippy::upper_case_acronyms` see https://rust-lang.github.io/rust-clippy/master/index.html#upper_case_acronyms

We can make an exception but don't think we should because this is a style choice that opposes the established convention.

Swiftb0y:
Thank you very much for the review. In the long-term we should really discuss error handling, but this is good enough for now IMO. I might still add some unit tests for the `Err` cases if you don't press merge in the meantime.

Swiftb0y:
addressed your review in a fixup commit. Tell me when this is ready then I'll squash everything before merge.

src/pdb/string.rs
```diff
@@ -0,0 +1,271 @@
...
+    },
+    /// Regular long form strings, containing possibly different encodings
+    Long {
+        #[br(temp)]
+        #[br(assert(flags & 0b1 == 0))]
+        #[bw(calc = content.flags())]
+        flags: u8,
+
+        #[br(temp)]
+        #[bw(calc = content.byte_count().unwrap() + 4)]
```
Holzhaus, line 154: This may panic... is there no way to return an error? I'm not sure if this is possible.

src/pdb/string.rs
```diff
@@ -0,0 +1,271 @@
...
+        #[bw(calc = (((content.len() + 1) << 1) | 1) as u8)]
+        header: u8,
+
+        #[br(count = (header >> 1) - 1)]
+        content: Vec<u8>,
+    },
+    /// Regular long form strings, containing possibly different encodings
+    Long {
+        #[br(temp)]
+        #[br(assert(flags & 0b1 == 0))]
```
Holzhaus, line 149: Why `& 0b1`? Shouldn't it be `flags == 0x40 || flags == 0x90`?

src/pdb/string.rs
Holzhaus: ok

src/pdb/string.rs
```diff
@@ -0,0 +1,271 @@
...
+//! `binrw`-based implementation for DeviceSQLStrings capable of parsing and
+//! serializing DeviceSQLStrings
+//!
+//! See <https://djl-analysis.deepsymmetry.org/rekordbox-export-analysis/exports.html#devicesql-strings>
+
+use binrw::{binrw, NullString};
+
+const MAX_SHORTSTR_SIZE: usize = ((u8::MAX >> 1) - 1) as usize;
+
+/// Error Objects occurring when dealing with [DeviceSQLString]'s
```
Holzhaus, line 18:
```suggestion
/// Error Objects occurring when dealing with [`DeviceSQLString`]s.
```

src/pdb/string.rs
```diff
@@ -0,0 +1,271 @@
...
+    /// Creating of ISRC String was unsuccessful because the string was not
+    /// containing a valid ISRC string specifier
+    InvalidISRC,
+    /// String encoding Error (invalid UTF-8/16)
+    Encoding,
+}
+
+/// Encapsulates the intrinsics of the format used by DeviceSQL strings
+///
+/// Once A deviceSQLstring has been constructed, there is no way to change it.
```
Holzhaus, line 33:
```suggestion
/// Once a `DeviceSQLString` has been constructed, there is no way to change it.
```

src/pdb/string.rs
```diff
@@ -0,0 +1,271 @@
...
+/// let parsed = DeviceSQLString::read(&mut reader)?;
+/// assert_eq!(parsed, string);
+/// # Ok(())
+/// # }
+/// ```
+#[derive(Debug, PartialEq)]
+#[binrw]
+pub struct DeviceSQLString(DeviceSQLStringImpl);
+impl DeviceSQLString {
+    /// Initializes a DeviceSQLString from a plain Rust String
```
Holzhaus, line 56:
```suggestion
    /// Initializes a [`DeviceSQLString`] from a plain Rust [`String`].
```

src/pdb/string.rs
```diff
@@ -0,0 +1,271 @@
...
+                Ok(Self(DeviceSQLStringImpl::Long {
+                    // note: The DeviceSQL database may only support UCS-2 so
+                    // we might need to do some additional filtering here
+                    content: LongBody::Ucs2le(string.encode_utf16().collect()),
+                }))
+            }
+        } else {
+            Err(StringError::TooLong)
+        }
+    }
```
Holzhaus, line 79:
```suggestion
    }

```

src/pdb/string.rs
```diff
@@ -0,0 +1,271 @@
...
+        }
+        // Basic validation taken from:
+        // https://isrc.ifpi.org/downloads/ISRC_Bulletin-2015-01.pdf
+        if string.len() != 12 || !string.is_ascii() {
+            return Err(StringError::InvalidISRC);
+        }
+        Ok(Self(DeviceSQLStringImpl::Long {
+            content: LongBody::Isrc(NullString::from_string(string)),
+        }))
+    }
```
Holzhaus, line 98:
```suggestion
    }

```

src/pdb/string.rs
```diff
@@ -0,0 +1,271 @@
...
+                content: LongBody::Isrc(str),
+                ..
+            } => str
+                .into_string_lossless()
+                .map_err(|_| StringError::Encoding),
+            DeviceSQLStringImpl::Long {
+                content: LongBody::Ucs2le(vec),
+            } => String::from_utf16(&vec).map_err(|_| StringError::Encoding),
+        }
+    }
```
Holzhaus, line 119:
```suggestion
    }

```

src/pdb/string.rs
```diff
@@ -0,0 +1,271 @@
...
+    /// Create an empty DeviceSQLString
```
Holzhaus, line 120:
```suggestion
    /// Create an empty [`DeviceSQLString`].
```

src/pdb/string.rs
```diff
@@ -0,0 +1,271 @@
...
+        } else {
+            Err(StringError::TooLong)
+        }
+    }
+    /// Creates a DeviceSQLString containing an ISRC instead of an expected string
+    ///
+    /// The DeviceSQL Database as used by rekordbox has a strange oddity
+    /// where it serializes strings containing a tracks ISRC (International
+    /// Standard Recording Code) in an unexpected format, if this is desired,
+    /// use this constructor function instead of [DeviceSQLString::new]
```
Holzhaus, line 85:
```suggestion
    /// use this constructor function instead of [`DeviceSQLString::new`].
```

src/pdb/string.rs
```diff
@@ -0,0 +1,271 @@
...
+    }
+
+    #[test]
+    fn short_ascii_string() -> Result<(), StringError> {
+        test_roundtrip(
+            &[0x9, 0x66, 0x6F, 0x6F],
+            DeviceSQLString::new("foo".to_owned())?,
+        );
+        Ok(())
+    }
```
Holzhaus, line 224: I'd also be okay with the more concise version, but I don't really care:

```suggestion
    #[test]
    fn short_ascii_string() {
        test_roundtrip(
            &[0x9, 0x66, 0x6F, 0x6F],
            DeviceSQLString::new("foo".to_owned()).unwrap(),
        );
    }
```

Holzhaus (commented):
Some more nitpicks, then I'll stop :D

src/pdb/string.rs
Swiftb0y: Its only going to return an error if the string we constructed was already too big in the first place (which we already check on construction). I also opted for this approach because I could not find another way of making this panic-less.

src/pdb/string.rs
Swiftb0y: Well, yes in the reference documentation but I recently gained some insight into DeviceSQLStrings that changed my view of them slightly. See the corresponding PR: https://github.com/Deep-Symmetry/crate-digger/pull/26/files

Swiftb0y: To summarize, the first byte serves doubly duty, it is either a 32-bit bitmask or an 8-bit bitmask. To differentiate between the two, the LSB of the first byte can be tested. This is what the asserts are doing here. In the 8-bit case, the upper 7 seven bits contain the length of the short ascii string and in the 32-bit case, the upper seven bits contain information about the contained string (probably encoding information). See [Greg Kemnitz blog post](https://bufferpool.quora.com/What-does-a-db-engine-string-look-like) for some more insights (greg is the author of the Encirq version of DeviceSQL). In the linked PR, I described a mapping of these bits to string metadata, but we can't be sure they are accurate them until we see other types of strings (which I think is unlikely).

Swiftb0y: So the only requirement for being a valid `Long` string, is `flags & 0b1 == 0`, the more indepth check is done in the `LongBody` enum were I `preassert` the flags to actually find the correct enum variant. This solution makes it easy to add new string variants if we ever encounter them.

Holzhaus: Could you add a comment with your explanations? That will make it easier to understand why we check the LSB.

Swiftb0y: Sure

Holzhaus:
Great, if you squash the commits we can merge now I think.

Swiftb0y:
Sorry for taking so long, I was fighting with merge conflicts during squashing.

Holzhaus (approved):

Swiftb0y:
Thank you

Holzhaus:
> Thank you

Thank *you*.

PR #49, opened 2022-02-25
setting/anlz: Fix field visibility and derive order

Holzhaus:


Swiftb0y:
what was your rationale for switching the attribute order?

I'm also split on the field visibility. On one hand, making unknown fields invisible makes sense from an API perspective on the other hand, having full visibility into the entire parsed buffer would make it possible to build tools for reverse engineering the meaning of those unknown fields.

Holzhaus:
> what was your rationale for switching the attribute order?

`#[br(temp)]` interferes with `#[derive(Debug)]` if the order is wrong. See https://github.com/jam1garner/binrw/pull/103.

Holzhaus:
> I'm also split on the field visibility. On one hand, making unknown fields invisible makes sense from an API perspective on the other hand, having full visibility into the entire parsed buffer would make it possible to build tools for reverse engineering the meaning of those unknown fields.

True. The names of these unknown fields could change at any time. Even the sizes could be wrong (e. g. one u32 fieldcould actually be two u16 fields). So as you mentioned it makes sense to hide the from the public API.

If someone wants to build custom tools to aid reverse engineering, we can revert this of course. In the meantime, we can use the binaries in this crate to inspect the debug representation of the structs including all hidden fields.

PR #52, opened 2022-02-25
setting: Calculate CRC16 checksum automatically

Holzhaus:
Follow-up to #50.

### DIFF
--- a/src/setting.rs
+++ b/src/setting.rs
@@ -17,11 +17,12 @@
 //!
 //! The exact format still has to be reverse-engineered.
 
-use binrw::{binrw, NullString};
+use binrw::{binrw, io::Cursor, BinWrite, Endian, NullString, WriteOptions};
 
 #[binrw]
 #[derive(Debug, PartialEq)]
 #[brw(little)]
+#[bw(import(no_checksum: bool))]
 /// Represents a setting file.
 pub struct Setting {
     /// Size of the string data field (should be always 96).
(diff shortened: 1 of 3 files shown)


src/setting.rs
```diff
@@ -17,11 +17,12 @@
 //!
 //! The exact format still has to be reverse-engineered.
 
-use binrw::{binrw, NullString};
+use binrw::{binrw, io::Cursor, BinWrite, Endian, NullString, WriteOptions};
 
 #[binrw]
 #[derive(Debug, PartialEq)]
 #[brw(little)]
+#[bw(import(no_checksum: bool))]
```
Swiftb0y, line 25: since this is basically a hack, I'd propose to make give it a indicative name and assign an appropriate default
```suggestion
#[bw(import(_no_checksum: bool = false))]
```

src/setting.rs
```diff
@@ -59,13 +60,43 @@ pub struct Setting {
...
+            // the section lengths and string data.
+            SettingData::DJMMySetting(_) => 0,
+            // In all other files`, the checksum is calculated just over the data section which
+            // starts at offset 104,
+            _ => 104,
+        };
+
+        let end = data.len() - 4;
+        crc16::State::<crc16::XMODEM>::calculate(&data[start..end])
+    }
```
Swiftb0y, line 97: instead of serializing the entire thing twice, an alternative approach could be to take a slice of the buffer

Swiftb0y:
I thought about this and also had another idea in mind. Would it be possible to impl `BinWrite` with an added `io::Read` bound? If we have guarantees that all previous fields have already been written, then calculating the checksum would be possible by seeking to the start again, reading everything up to the checksum field and then calculating the checksum from that. Not sure if that is possible / reliable though but thought I'd mention it in case you'd find it interesting enough to look into it.

Holzhaus:
> Would it be possible to impl `BinWrite` with an added `io::Read` bound?

I don't think so. Streams could potentially be write-only.

Swiftb0y:
I don't think so. Streams could potentially be write-only, e.g. network streams. I'd consider such a restriction worse than the additional serialization cost (considering that these setting files are quite small).

I'm aware, but I find it very unlikely that this will be used in conjunction with a write-only sink.

> I'd consider such a restriction worse than the additional serialization cost (considering that these setting files are quite small).

I agree, I was not proposing to replace the double-serialization but rather to specialize for the case where `Read` is available as well (so its more an optimization). Not sure if this is possible with Rust though.

src/setting.rs
Holzhaus: I don't think the default argument is possible. And I don't understanrd the purpose of the underscore.

Swiftb0y: My current understanding is that the `no_checksum` argument is basically a hack so we can serialize the `Setting` struct once without a checksum and once with. This is a workaround because of lacking support to do this "properly" with binrw. As such, it is an implementation detail. If I just want to serialize the `Setting` struct, I don't want to worry about this internal hack. As such, I propose to make the `no_checksum` attribute optional by using an appropriate default argument and indicating that its private (by prefixing it with `_` per convention). 

defaults args are supposed to be possible per documentation (see the example code and single sentence https://docs.rs/binrw/latest/binrw/attribute/read/index.html#named-arguments).

Holzhaus: Ah, but then import needs a struct, not a tuple.

Holzhaus: Ah, with named arguments a struct is generated and we'd have to reference its name (which is an implementation detail of binrw) in the `write_options` implementation :(
It also interferes with the `missing_docs` clippy rule.

And the leading underscore prevents any unused-variable warnings, doesn't it?

Swiftb0y: > Ah, with named arguments a struct is generated and we'd have to reference its name (which is an implementation detail of binrw) in the write_options implementation :(

Mhmm. that sounds like an issue worth raising on the binrw bugtracker. 

> And the leading underscore prevents any unused-variable warnings, doesn't it?

true, I forgot about that convention. 

I really wish there was a solution that didn't expose the double-serialization workaround in the public interface, but I guess there is none.

Holzhaus: > I really wish there was a solution that didn't expose the double-serialization workaround in the public interface, but I guess there is none.

Yeah. Maybe we can come up with a better implementation later on.

PR #53, opened 2022-02-26
anlz: Add support for XOR-encrypted song structure sections

### LINKED ISSUE #13: anlz: Add support for XOR-encrypted Song Structure Tags (rekordbox 6+)
From the [documentation](https://djl-analysis.deepsymmetry.org/rekordbox-export-analysis/anlz.html#song-structure-tag):

> The version that rekordbox 6 exports is garbled with an XOR mask to make it more difficult to access the data. All bytes after lene (bytes 10-11) are XOR-masked with a pattern that is generated by adding the value of lene to each byte of the following base pattern:
>
>     CB E1 EE FA E5 EE AD EE E9 D2 E9 EB E1 E9 F3 E8 E9 F4 E1

Support for this should be added.

Holzhaus:
Fixes #13.

### DIFF
--- a/src/xor.rs
+++ b/src/xor.rs
@@ -1,0 +1,88 @@
+// Copyright (c) 2022 Jan Holthuis
+//
+// This Source Code Form is subject to the terms of the Mozilla Public License, v. 2.0. If a copy
+// of the MPL was not distributed with this file, You can obtain one at
+// http://mozilla.org/MPL/2.0/.
+//
+// SPDX-License-Identifier: MPL-2.0
+
+//! Wrapper types for dealing with XOR encryption.
+
+use binrw::io::{Read, Result, Seek, SeekFrom, Write};
+use std::iter::Cycle;
+
+/// Stream cipher wrapper around another [`Read`]- or [`Write`]-able stream that XOR's all bytes
+/// after reading/before writing.
+#[derive(Debug)]
+pub struct XorStream<T> {
+    stream: T,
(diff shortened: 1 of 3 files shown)


src/xor.rs
```diff
@@ -0,0 +1,88 @@
...
+use binrw::io::{Read, Result, Seek, SeekFrom, Write};
+use std::iter::Cycle;
+
+/// Stream cipher wrapper around another [`Read`]- or [`Write`]-able stream that XOR's all bytes
+/// after reading/before writing.
+#[derive(Debug)]
+pub struct XorStream<T> {
+    stream: T,
+    key: Cycle<std::vec::IntoIter<u8>>,
+    key_size: u64,
```
Swiftb0y, line 20: since there is no way to create a key that is larger than `usize`, I don't understand why the key_size is `u64`

Swiftb0y (commented):
very minor nitpick. Everything else looks good. Kinda awkward at times but I can't think of a better implementation myself either.

src/xor.rs
Holzhaus: Because we only need the `key_size` for the `Seek` trait, and [`seek()` uses `u64`](https://doc.rust-lang.org/std/io/trait.Seek.html#tymethod.seek). We could store a `usize`, but then we'd have to convert each time seek is called.

Swiftb0y (approved):

PR #56, opened 2022-03-09
pre-commit: Add hook to add sourceheaders automatically

Holzhaus:


.pre-commit-config.yaml
```diff
@@ -50,3 +50,7 @@ repos:
       -D,
       warnings,
     ]
+- repo: https://github.com/Holzhaus/sourceheaders
+  rev: 41391140ca6ac7c3dfddae090353d183cbeadb48
+  hooks:
+  - id: sourceheaders
```
Swiftb0y, line 56: would've wished this was written in Rust (cuz u know, [oxidize it](https://wiki.mozilla.org/Oxidation)) but python makes sense too for pre-commit. ;)

build.rs
```diff
@@ -1,3 +1,10 @@
+// Copyright (c) 2022 Jan Holthuis <jan.holthuis@rub.de>
+//
+// This Source Code Form is subject to the terms of the Mozilla Public License, v. 2.0. If a copy
+// of the MPL was not distributed with this file, You can obtain one at
+// http://mozilla.org/MPL/2.0/.
+//
+// SPDX-License-Identifier: MPL-2.0
 use glob::glob;
```
Swiftb0y, line 8: I think it would be a nice addition to `sourceheaders` if it put at least one newline after the license header.

Swiftb0y: Or you can just do that yourself I guess since its purely a style choice.

.pre-commit-config.yaml
Holzhaus: I thought about it, but unfortunately rust code requires a system install of rust on the host, so I decided against it.

build.rs
Holzhaus: Done.

Swiftb0y (approved):

PR #58, opened 2022-03-15
Add gitlint hook to pre-commit configuration

Holzhaus:


Swiftb0y (approved):

Swiftb0y:
Thanks, We need a file where we document that...

PR #64, opened 2022-04-03
Add binrw support for pdb files

### LINKED ISSUE #45: Port `pdb.rs` from nom to binrw
@Swiftb0y is working on a PoC: https://github.com/Holzhaus/rekordcrate/issues/39#issuecomment-1049160961

Holzhaus:
Fixes #45. Based on #66 and #67.


Holzhaus:
I think this is ready to review now.

src/pdb/mod.rs
```diff
@@ -325,50 +255,62 @@ pub struct Page {
...
     unknown7: u16,
+
+    #[br(temp)]
+    #[bw(ignore)]
+    #[br(calc = if num_rows_large > num_rows_small.into() && num_rows_large != 0x1fff { num_rows_large } else { num_rows_small.into() })]
+    num_rows: u16,
+
+    #[br(temp)]
+    #[bw(ignore)]
+    #[br(calc = if num_rows > 0 { (num_rows - 1) / RowGroup::MAX_ROW_COUNT + 1 } else { 0 })]
```
Swiftb0y, line 266: unfortunately this is still nightly-only, but I think we should leave this in the code so we can go back and replace it eventually once its been stabilized.
```suggestion
    #[br(calc = num_rows.div_ceil(RowGroup::MAX_ROW_COUNT))]
```

src/pdb/mod.rs
```diff
@@ -701,350 +603,70 @@ pub enum Row {
         /// Unknown field (alternating "2" and "3"?).
         unknown7: u16,
         /// International Standard Recording Code (ISRC), in mangled format.
+        #[br(offset = base_offset, parse_with = FilePtr16::parse)]
         isrc: DeviceSQLString,
```
Swiftb0y, line 607: why did you opt for the `parse_with` approach instead?
```suggestion
        #[br(offset = base_offset)]
        isrc: FilePtr16<DeviceSQLString>,
```

src/pdb/mod.rs
```diff
@@ -325,50 +255,62 @@ pub struct Page {
...
+
+        // Read last row group
+        let row_group = RowGroup::read_options(reader, ro, (num_rows_in_last_row_group,))?;
+        row_groups.push(row_group);
+
+        // Read remaining row groups
+        for _ in 1..num_row_groups {
+            let row_group = RowGroup::read_options(reader, ro, (RowGroup::MAX_ROW_COUNT,))?;
+            row_groups.insert(0, row_group);
+        }
```
Swiftb0y, line 311: I dislike the fact that this is kinda brittle in regards to ordering and so forth. Also the repeated insertion at the front is not ideal but I don't see a good alternative either.

src/bin/rekordcrate-pdb.rs
```diff
@@ -6,34 +6,48 @@
...
+                    reader
+                        .seek(SeekFrom::Start(abs_offset))
+                        .expect("failed to seek to row offset");
+                    let row = Row::read_options(
+                        &mut reader,
+                        &ReadOptions::default(),
+                        (page.page_type.clone(),),
+                    )
+                    .expect("failed to parse row");
+                    println!("      {:?}", row);
```
Swiftb0y, line 48: having to manually mess with the offset is highly unsafe and not something I would make part of the API in any shape. First of all because a lot could go wrong when the consumer is manually in charge of juggling all these types, and the resulting boilerplate required to make this work is also less than ideal. Basically, its easy to misuse and hard to use correctly. https://www.oreilly.com/library/view/97-things-every/9780596809515/ch55.html

Swiftb0y: Also, making binrw part of the public API is not good either.

Swiftb0y (changes requested):
Thanks for pinging me that this is ready. 
General issue still: We use the `binrw` attribute everywhere, but writing obviously produces garbage databases (even though thats not obvious from the code). So I think we should either make that clear at compile time (by using `#[binread]` on the read-only datastructures) or if that requires too many changes, document that writing won't produce valid databases!

src/util.rs
```diff
@@ -54,14 +45,6 @@ pub enum ColorIndex {
     Unknown(u16),
```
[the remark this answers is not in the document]
Holzhaus: good catch, but this code was actually still used for `nom`.

85567a05f2f392336bbef0bd6186ee317f93f013

src/pdb/mod.rs
```diff
@@ -201,64 +135,60 @@ pub struct Header {
...
+    pub fn read_pages<R: Read + Seek>(
+        &self,
+        reader: &mut R,
+        ro: &ReadOptions,
+        args: (&PageIndex, &PageIndex),
+    ) -> BinResult<Vec<Page>> {
+        let (first_page, last_page) = args;
+
+        let mut pages = vec![];
+        let mut page_index = first_page.clone();
```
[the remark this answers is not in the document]
Holzhaus: `pages.push(page);` is a move, which means we cannot use the borrowed `page.next_page` in the next iteration of the loop, which is why we need to clone.

The line you pointed out the clone is only necessary because there would be a type mismatch between `&PageIndex` and `PageIndex`.

src/pdb/mod.rs
```diff
@@ -201,64 +135,60 @@ pub struct Header {
...
+        loop {
+            let page_offset = self.page_offset(&page_index);
+            reader
+                .seek(SeekFrom::Start(page_offset))
+                .expect("failed to seek to page offset");
```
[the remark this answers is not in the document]
Holzhaus: Sorry, the expect is a leftover from initial experimentation.

src/pdb/mod.rs
Holzhaus: Because the actual offset is an implementation detail IMHO. When we add support for writing, I don't think we want to save the offsets in the struct, otherwise all later offsets of a row become wrong when an earlier string changes length.

src/pdb/mod.rs
Holzhaus: I contemplated if I should use a `VecDeque` instead, but then decided against it. Didn't want to spend too much time on premature optimization, we can always fine-tune the code later.

src/bin/rekordcrate-pdb.rs
Holzhaus: > having to manually mess with the offset is highly unsafe and not something I would make part of the API in any shape. First of all because a lot could go wrong when the consumer is manually in charge of juggling all these types, and the resulting boilerplate required to make this work is also less than ideal. Basically, its easy to misuse and hard to use correctly. https://www.oreilly.com/library/view/97-things-every/9780596809515/ch55.html

I agree and already have some code that gets rid of this. The downside is that it just parse all rows in a page at once, not lazily.

> Also, making binrw part of the public API is not good either.

There is only alternative I can think of (other than to not use binrw at all) would be to copy all parsed data into another struct that does not implement `BinRead`/`BinWrite`. If the `trait` and the `struct` is `pub`, so is the `impl`. We can document that using `binrw` is an implementation detail and depending on it is not recommended though.

Anyway, I wouldn't worry too much about the public API yet. We are still in the experimentation phase, and hiding binrw stuff just makes the code complicated without much gain.  Later on, we need to think how to make serialization possible without in-depth knowledge of the format without creating struct instances by hand anyway. Whoever uses this library probably doesn't care about the various string types, page size and how many rows there are in a row group.

Holzhaus:
Thanks for reviewing!

> General issue still: We use the `binrw` attribute everywhere, but writing obviously produces garbage databases (even though thats not obvious from the code). So I think we should either make that clear at compile time (by using `#[binread]` on the read-only datastructures) or if that requires too many changes, document that writing won't produce valid databases!

Agreed. Fixed in https://github.com/Holzhaus/rekordcrate/pull/64/commits/076efb03f6b0ba3669b191f68b7c9942954e3810.

src/pdb/mod.rs
Holzhaus: https://github.com/Holzhaus/rekordcrate/pull/64/commits/359db9d84ef7a9b40c1e0299ed19a8dadec11dd0

src/pdb/mod.rs
```diff
@@ -368,9 +373,9 @@ pub struct RowGroup {
     rows: Vec<RowOffset>,
     row_presence_flags: u16,
     /// Unknown field, probably padding.
-    #[br(temp)]
-    #[bw(calc = 0)]
-    padding: u16,
+    ///
+    /// Apparently this is not always zero, so it might also be something different.
+    unknown: u16,
```
Swiftb0y, line 378: can we still assert that its zero so we can more easily catch occurrences of this?

src/pdb/mod.rs
```diff
@@ -539,8 +555,10 @@ pub enum Row {
     /// Contains the album name, along with an ID of the corresponding artist.
     #[br(pre_assert(page_type == PageType::Tracks))]
     Track {
+        /// Position of start of this row (needed of offset calculations).
+        ///
+        /// **Note:** This is a virtual field and not actually read from the file.
         #[br(temp, parse_with = current_offset)]
-        #[bw(ignore)]
```
Swiftb0y: what does the removal of `#[bw(ignore)]` accomplish? The struct is still writable, its just that the resulting binary is even more "wrong". 
IMO we need to take very special care with these "virtual fields" and make sure they don't end up influencing the reader/writer, otherwise they will not work correctly.

src/pdb/mod.rs
```diff
@@ -259,17 +262,14 @@ pub struct Page {
     unknown7: u16,
 
     #[br(temp)]
-    #[bw(ignore)]
```
Swiftb0y: I would've preferred to comment them out instead of removing them. Otherwise we might not know what to add back when we actually do add writing capability.

Swiftb0y (changes requested):

src/pdb/mod.rs
Swiftb0y: Mhmm, that makes sense, but since `FilePtr` does have BinWrite yet, it doesn't make much sense to discuss the ergonomics of a foreign, still unimplemented API.

src/bin/rekordcrate-pdb.rs
Swiftb0y: Don't get me wrong, I agree that completely eliminating binrw from our public API is not worth the effort, but the amount of boilerplate and types needed for the basic act of just reading out all rows is not good. If we only require the API consumer to make use of BinRead/BinWrite, thats fine IMO. 

> I agree and already have some code that gets rid of this. The downside is that it just parse all rows in a page at once, not lazily. Didn't want to put it into this PR because it's already a pretty large diff.

Ok, then I won't dwell on this issue in this PR any longer.

src/pdb/mod.rs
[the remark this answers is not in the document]
Swiftb0y: Makes sense, thanks for pointing that out. All hail borrowck ;)

src/pdb/mod.rs
Holzhaus: > The struct is still writable, its just that the resulting binary is even more "wrong".

It's not though? The enum struct is binread, not binrw.

src/pdb/mod.rs
Holzhaus: `br(temp)` fields throw a compiler warning when the struct is `#[binrw]` and we forget to add back `bw(ignore)`:

```
error: The field num_rows is temp but not provided a value in its BinWrite derivation. Try using `bw(calc = ...)` or `bw(ignore)`
   --> src/pdb/mod.rs:259:5
    |
259 |     /// Number of rows in this page.
    |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    ```

File: src/pdb/mod.rs (hunk 8 above)
Comment (Holzhaus) [reply src/pdb/mod.rs:378]: No, that makes tests fail:

```
thread 'pdb_demo_tracks_PIONEER_rekordbox_export_pdb' panicked at 'called `Result::unwrap()` on an `Err` value:
 ╺━━━━━━━━━━━━━━━━━━━━┅ Backtrace ┅━━━━━━━━━━━━━━━━━━━━╸

 0: Error: unknown == 0 at 0x2fee
           While parsing field 'row_groups' in Page
     at src/pdb/mod.rs:282

 ╺━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━╸

', /home/jan/Projects/rekordcrate/target/debug/build/rekordcrate-2debded95e6c6d26/out/tests_pdb.rs:23:14
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

src/bin/rekordcrate-pdb.rs
Holzhaus: Here's the commit btw: https://github.com/Holzhaus/rekordcrate/commit/88c87c83f87ede77adb8fbe4fcb8ec46c06613b7

src/pdb/mod.rs
Swiftb0y: I didn't know that. Then ignore my previous comment.

src/pdb/mod.rs
Swiftb0y: Mhmm, I thought this might not be that common. Then I guess we leave the assert out and try to decipher the meaning of these fields some other time.

Swiftb0y: Mhmm, I thought this might not be that common. Then I guess we leave the assert out and try to decipher the meaning of these fields some other time.

Swiftb0y (approved):

PR #65, opened 2022-04-07
Remove wrong `unknown8` field from `Page` struct

Holzhaus:
The field actually belongs to each row group rather than the page.
This change is a leftover from commit
2e2afd2a7f9feb4842289805eaec228e5bb166c0 which fixed the offset
calculation but forgot to remove the corresponding field in the `Page`
struct.

Also see these PRs for details:

- https://github.com/Holzhaus/rekordcrate/pull/20
- https://github.com/Deep-Symmetry/crate-digger/pull/23


Swiftb0y (approved):
Thanks.

PR #66, opened 2022-04-07
Use 8-bit color indices consistently

Holzhaus:
Before, we sometimes used 8-bit indices and sometimes 16-bit ones,
because the reverse-engineered format documentation described it that
way.

However, these does not seem to be a good reason why there should be
16-bit color indices when the value it contains is just 8-bit. It's
likely that the most significant byte actually belongs to another field.

Also see:

  https://github.com/Deep-Symmetry/crate-digger/issues/27

### DIFF
--- a/src/pdb/mod.rs
+++ b/src/pdb/mod.rs
@@ -629,7 +629,7 @@ pub enum Row {
         /// Numeric color ID
         color: ColorIndex,
         /// Unknown field.
-        unknown3: u8,
+        unknown3: u16,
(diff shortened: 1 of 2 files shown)


Swiftb0y (approved):
Code looks good to me. Though we should wait with merging until this was discussed "upstream" (https://github.com/Deep-Symmetry/crate-digger/issues/27)

Swiftb0y (approved):
I guess there is no upstream controversy regarding the change, so LGTM

PR #67, opened 2022-04-08
DeviceSQLString Unification

Holzhaus:
Merge both DeviceSQLString implementations and improve debug output.

### DIFF
--- a/src/pdb/mod.rs
+++ b/src/pdb/mod.rs
@@ -20,6 +20,7 @@
 
 pub mod string;
 
+use crate::pdb::string::DeviceSQLString;
 use crate::util::{nom_input_error_with_kind, ColorIndex};
 use nom::error::ErrorKind;
 use nom::IResult;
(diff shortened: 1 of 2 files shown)


src/pdb/mod.rs
```diff
@@ -495,81 +496,20 @@ impl RowGroup {
...
-    LongASCII(String),
-    /// A long UTF-16 little-endian encoded string (max 32767 characters).
-    LongUTF16LE(String),
-    /// A ShortASCII variant that is needed to workaround the special ISRC string
-    /// NOTE: the serialized string is a null-terminated C-String
-    ISRCASCII(String),
-}
-
 impl DeviceSQLString {
     fn parse(input: &[u8]) -> IResult<&[u8], DeviceSQLString> {
```
Swiftb0y, line 500: I'm not sure I understand. This basically adds another impl with a nom-compatible parse method, so we can use the binrw-based DeviceSQLString type in the context of nom?

Holzhaus: Correct. This allows removing the old `DeviceSQLImplementation` completely.

This `impl` block is removed in #64, because it's only needed for the transitional phase where the binrw parser still needs nom.

Holzhaus: Any objections?

Swiftb0y: Nope, thanks for the explanation. I was confused by the disconnected impl block, but if it compiles, it works.

Swiftb0y (approved):
Thanks LGTM

src/pdb/mod.rs
Holzhaus: Yeah, the `string.rs` module is currently binrw-only, and I wanted to keep it that way, this is why I put the impl block here (which is also the only module where the method is actually used).

Issue #68, opened 2022-04-08
Serialization Support for PDB files


Holzhaus:
Basic serialization is already supported for analysis/setting files, but not PDB files. When #64 has been merged, we can look into implementing support for this.

Note that the PDB parser uses [`FilePtr`](https://docs.rs/binrw/latest/binrw/struct.FilePtr.html) which does not support serialization out of the box yet (see https://github.com/jam1garner/binrw/issues/4), so we'll have to provide a custom implementation.

lukaprincic:
Thanks everyone for working on this. Holding fingers crossed 🤞🏼 for this, I use MIXXX exclusively on Linux a lot but sometimes I need to use CDJs and it's a pain to go to proprietary systems to prepare USB keys for the CDJs.

ambientsound:
You're welcome! I haven't contributed to this project, but I've been doing some more work on reverse engineering PDB exports, to the point where I can actually generate a PDB file that I can open and use in Rekordbox. I haven't tested it on hardware yet though. If my work can be helpful to this project I would be happy to share my findings.

Holzhaus:
@ambientsound hi, did you look into the inofficial documentation of the format? https://djl-analysis.deepsymmetry.org/rekordbox-export-analysis/exports.html

If you found something while reversing  that is not yet documented, it would be nice to [add it to the docs](https://github.com/Deep-Symmetry/crate-digger/tree/main/doc).

ambientsound:
@Holzhaus I found some new stuff. But I struggled with Kaitai, and found the generated code detrimental to the progress of the project. But there is a bunch of documentation in my type structs and code, there should be something useful.

It's not really ready to publish, but I tried to clean it up a bit and did it anyway so you can have a look.

https://github.com/ambientsound/rex

ambientsound:
Short summary of what I've done:

Most of what I figured out revolves around the "strange pages", which I believe to be some kind of table index, together with flags for this particular table. I found the data structure but haven't figured out all the values.

Also, I found the meaning of a few more fields scattered around different tables, and extracted some information about the default data exported in a pristine library with nothing in it.

There is a command-line tool that should be able to generate export files for further testing.

Hope it helps, I'm also open to suggestions or questions.

lukaprincic:
Apologies to ping this, but I was wondering if there's any progress on this?

ambientsound:
Hi @lukaprincic 
From my part there is no more progress than what is happening in the rex repo.
[...]

PR #71, opened 2022-04-23
PDB Serialization (Pt. 2)

Holzhaus:
Currently not working correctly, as the unittest fails. I'm currently a bit busy and don't have too much time to debug it. If anyone else wants to start working on it or give feedback, I'd appreciate it.

Related issue: #68

### DIFF
--- a/src/pdb/mod.rs
+++ b/src/pdb/mod.rs
@@ -478,9 +478,9 @@ pub struct PlaylistTreeNodeId(pub u32);
 pub struct HistoryPlaylistId(pub u32);
 
 /// Contains the album name, along with an ID of the corresponding artist.
-#[binread]
+#[binrw]
 #[derive(Debug, PartialEq, Eq, Clone)]
-#[br(little)]
+#[brw(little)]
 pub struct Album {
     /// Position of start of this row (needed of offset calculations).
     ///
@@ -932,9 +932,9 @@ impl BinWrite for Track {
 }
 
 /// A table row contains the actual data.
-#[binread]
+#[binrw]
 #[derive(Debug, PartialEq, Eq, Clone)]
-#[br(little)]
+#[brw(little)]
 #[br(import(page_type: PageType))]
 // The large enum size is unfortunate, but since users of this library will probably use iterators
 // to consume the results on demand, we can live with this. The alternative of using a `Box` would
--- a/src/pdb/mod.rs
+++ b/src/pdb/mod.rs
@@ -426,55 +426,55 @@ impl RowGroup {
 /// Identifies a track.
 #[binrw]
 #[derive(Debug, PartialEq, Eq, Clone, Copy, Hash)]
-#[br(little)]
+#[brw(little)]
 pub struct TrackId(pub u32);
 
 /// Identifies an artwork item.
 #[binrw]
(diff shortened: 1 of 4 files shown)


src/pdb/mod.rs
```diff
@@ -1184,4 +1304,634 @@ mod test {
...
+                ],
+                row_presence_flags: 96,
+                unknown: 64,
+            }],
+        };
+
+        let page_size: u32 = 4096;
+        test_roundtrip_with_args(
+            &[
+                0, 0, 0, 0, 2, 0, 0, 0, 0, 0, 0, 0, 47, 0, 0, 0, 32, 0, 0, 0, 0, 0, 0, 0, 7, 64, 0,
```
Swiftb0y, line 1760: Where did you get this from? Just exporting from rekordbox?

Swiftb0y (commented):
I'll try to debug why this doesn't work. I hope you don't mind that I use the review function to ask questions about specific parts of the code.

src/pdb/mod.rs
Holzhaus: Yes, IIRC It's a page from the rekordbox export in this repo. I should've documented this in a comment.

Holzhaus:
> I'll try to debug why this doesn't work. I hope you don't mind that I use the review function to ask questions about specific parts of the code.

Don't worrry and thanks for looking into this.

Swiftb0y:
I think I've gotten closer to the solution. One problem is that the export we generate has the rows in reverse. The other one is that rekordbox has additional padding between each row. I'll try to find a reason or a pattern to emulate.

src/pdb/mod.rs
Swiftb0y: I'll add a comment when I open the PR with the solution.

src/pdb/mod.rs
```diff
@@ -381,8 +450,8 @@ pub struct RowGroup {
     /// An offset which points to a row in the table, whose actual presence is controlled by one of the
     /// bits in `row_present_flags`. This instance allows the row itself to be lazily loaded, unless it
     /// is not present, in which case there is no content to be loaded.
-    #[br(offset = page_heap_offset, args { count: num_rows.into(), inner: (page_type,) })]
-    rows: Vec<FilePtr16<Row>>,
+    #[br(offset = page_heap_offset, parse_with = FilePtr16::parse, args { count: num_rows.into(), inner: (page_type,) })]
+    rows: Vec<Row>,
```
Swiftb0y, line 454: This is not the same and thus this branch also broke plain reading unfortunately. The proper implementation is a vector of pointers while the new one is a pointer of a vector. 

Also, since the manual write implementation `RowGroup::write_options_and_get_row_offset` is already quite complex I'm questioning it makes sense to continue down this path. In the end it comes down to requiring some sort of `FilePtr` serialization IMO. Also, some sort of abstract heap would be nice too (since Pages are also essentially just heap allocated too).

Holzhaus: Well,  that would require computing the offsets beforehand (to instantiate `FilePtr32`), which would make it hard to create a PDB file from scratch. The user likely does not care about the offsets. This is why I chose this approach, but I'm open to ideas.

Holzhaus:
Remove the broken code so that we can get the working changes in.

PR #75, opened 2022-09-07
refactor(setting): enum `#[derive(Default)]` instead of `impl Default`

Swiftb0y:
https://doc.rust-lang.org/std/default/trait.Default.html#enums


Holzhaus (approved):
Nice, this is much better. Thank you.

PR #78, opened 2022-09-29
refactor: update to `binrw` 0.10

Swiftb0y:
unfortunately in order to keep CI green, a bunch of things were updated in parallel. (making cargo fmt, clippy, etc happy while also introducing the changes necessary for binrw 0.10). 

If you insist, I can try to reorder and split some of the changes.

### DIFF
--- a/src/anlz.rs
+++ b/src/anlz.rs
@@ -226,7 +226,7 @@ pub struct Cue {
 
 /// A memory or hot cue (or loop).
 #[binrw]
-#[derive(Debug, PartialEq)]
+#[derive(Debug, PartialEq, Eq)]
 #[brw(big)]
 pub struct ExtendedCue {
     /// Cue entry header.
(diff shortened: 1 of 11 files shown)


src/anlz.rs
```diff
@@ -525,58 +525,58 @@ pub struct Phrase {
 
 /// Section content which differs depending on the section type.
 #[binrw]
-#[derive(Debug, PartialEq)]
+#[derive(Debug, PartialEq, Eq)]
 #[br(import(header: Header))]
 pub enum Content {
     /// All beats in the track.
-    #[brw(pre_assert(header.kind == ContentKind::BeatGrid))]
+    #[br(pre_assert(header.kind == ContentKind::BeatGrid))]
```
Holzhaus, line 532: Does `pre_assert` only work with `br`? I used `brw` on purpose to prevent writing inconsistent data, but maybe it never worked anyway...

Holzhaus (commented):
Thanks.

Swiftb0y:
Thanks :heart:

PR #80, opened 2022-09-29
Merge main into pdb pt2

Swiftb0y:
CI is obviously still broken but at least this branch is now up to date and warning-free again so somebody can work on fixing it.

### DIFF
--- a/src/pdb/mod.rs
+++ b/src/pdb/mod.rs
@@ -186,9 +186,6 @@ impl Header {
 /// Each page consists of a header that contains information about the type, number of rows, etc.,
 /// followed by the data section that holds the row data. Each row needs to be located using an
 /// offset found in the page footer at the end of the page.
-///
-/// **Note: The `Page` struct is currently not writable, because row offsets are not taken into
-/// account and rows are not serialized correctly yet.**
 #[binread]
 #[derive(Debug, PartialEq)]
 #[br(little, magic = 0u32)]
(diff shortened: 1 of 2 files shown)


Holzhaus:
Thanks, but the branch contains unfinished WIP commits that I don't want to end up in the git history of the main branch, so I think merging main is not the best option here. I rebased the branch on latest main instead.

Swiftb0y:
Can you elaborate? this PR did exactly that. Merging `main` into `pdb-serialization-pt2`, not the other way around. I issued this in PR form since I had to resolve some merge conflicts and I obviously also don't have the access to push to your branch.

Holzhaus:
It's basically impossible to amend the "WIP" commit (6a7e25d) once this PR would have been merged (due to the merge commit 3e89dd6 that comes after), which means there would be a broken WIP commit in `main`'s git history if we ever merge the `pdb-serialization-pt2` branch into `main` at some point.

Swiftb0y:
Ok that makes sense. Thanks for taking care of rebasing then. Just out of curiosity: Why wouldn't fixups of the WIP commits not work?

Holzhaus:
Rebasing with merge commits is messy because they have two parents. It might be possible but any time I tired it I found it cumbersome and just rebasing turned out the better option ;-)

Swiftb0y:
Great. thanks for the wisdom. :+1:

Holzhaus:
> Great. thanks for the wisdom. +1

`s/wisdom/incompetence when dealing with merge commits/` :wink:

PR #85, opened 2022-10-06
chore(pdb): Implement `Eq` trait on all row/string structs

Holzhaus:


Swiftb0y (approved):
LGTM, though I have couple thoughts:
Right now, we just derive `PartialEq` on most (all?) structs which implies `Eq` should hold as well. If we decide later that `PartialEq` should only compare fields whose meaning we have reverse-engineered (for a more pleasant API for example), the reflexivity would not hold anymore thus `Eq` would have to be removed, resulting in breaking the API.

Holzhaus:
> LGTM, though I have couple thoughts: Right now, we just derive `PartialEq` on most (all?) structs which implies `Eq` should hold as well. If we decide later that `PartialEq` should only compare fields whose meaning we have reverse-engineered (for a more pleasant API for example), the reflexivity would not hold anymore thus `Eq` would have to be removed, resulting in breaking the API.

Yup, but I don't really worry about that yet. It's still too early for a stable API IMHO. At the moment, almost any change is an API change (e.g., #84).

Side note: Ideally, at some point when the basic PDB parsing and serialization is done, we add a more high-level abstraction for the format, so that user's don't need to worry about the intricacies of the PDB format, and just use it like `Library::new("/run/media/foo/bar/").iter_albums().for_each(|album| println!("{} - {}", &album.artist().name(), &album.name()`) or something like that.

PR #86, opened 2022-10-07
Add C FFI [WIP]

Holzhaus:


Swiftb0y:
Is there a specific reason to prefer c bindings using cbindgen over c++ binding using cxx?

PR #87, opened 2022-10-07
build(cargo): Bump version to 0.2.0

Holzhaus:
Since v0.1.0 we switched from `nom` to `binrw` and added support for parsing `*SETTING.DAT` files, so it's really more than just a bug fix release. Although the switch to `binrw` is an API change, the major we don't want to bump the version `1` in order to emphasize that the API is not finalized and subject to changes.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "rekordcrate"
-version = "0.1.1"
+version = "0.2.0"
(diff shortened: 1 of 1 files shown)


Holzhaus:
@Swiftb0y might make sense to make a new release on crates.io. The current version is quite old already. What do you think?

Swiftb0y:
I agree with both (bumping to 0.2.0 and creating a new release).

Swiftb0y (approved):

PR #88, opened 2022-10-10
chore(changelog): update action versions

Swiftb0y:
Fixing this warning. 
Screenshot from 2022-10-10 19-28-30

Updated `EndBug/add-and-commit` as well. I removed the manual token passing. According to [a section in the Changelog](https://github.com/EndBug/add-and-commit/blob/90fb64e03a6338b2495d967c36266e0639dd104e/CHANGELOG.md#700---2021-01-16) the action will use the GITHUB_ by default.


Holzhaus:
Thanks. `ci` is more fitting than `chore` IMO though.

PR #91, opened 2022-10-11
Errors

Holzhaus:


src/util.rs
```diff
@@ -9,6 +9,19 @@
 //! Common types used in multiple modules.
 
 use binrw::binrw;
+use thiserror::Error;
+
+///  enumerates all possible errors returned by this library.
+#[derive(Error, Debug)]
+pub enum RekordcrateError {
```
Swiftb0y, line 16: So adding more variants later is not a breaking change.
```suggestion
#[non_exhaustive]
pub enum RekordcrateError {
```

Swiftb0y:
also `chore(util)` is not really correct for adding error handling imo. should be a feature instead. IIRC `chore` won't even show up in the changelog.

Holzhaus:
This is intentional, because it does not actually add the error handling (only in `main` where it doesn't really make a big difference). It just adds the necessary types for it. Once we actually add error handling to the library code, it should show up on the changelog.

Swiftb0y (approved):

Swiftb0y (approved):

Swiftb0y (approved):

PR #92, opened 2022-10-11
High Level API

Holzhaus:
This is an attempt to look into a high-level API to make it easier to read from (and later also write to) Rekordbox device exports without knowing much about the internals.

As first step, I only added support for setting files:

```rust
let mut export = DeviceExport::default();
export.load(path)?;
let settings = export.get_settings();
````

The CLI makes use of the new API when calling the new `list-settings` subcommand:

```
$ rekordcrate list-settings data/complete_export/demo_tracks/
On Air Display:                 On
LCD Brightness:                 3
Quantize:                       On
Auto Cue Level:                 Memory
Language:                       English
Jog Ring Brightness:            2 (Bright)
Jog Ring Indicator:             On
Slip Flashing:                  On
Disc Slot Illumination:         2 (Bright)
Eject Lock:                     Unlock
Sync:                           Off
Play Mode:                      Single / Off
Quantize Beat Value:            1 Beat
Hotcue Autoload:                On
Hotcue Color:                   Off
Needle Lock:                    Lock
Time Mode:                      Remain
Jog Mode:                       Vinyl
Auto Cue:                       On
Master Tempo:                   Off
Tempo Range:                    ±10%
Phase Meter:                    Type 1
Vinyl Speed Adjust:             Touch
Jog Display Mode:               Auto
Pad Button Brightness:          3
Jog LCD Brightness:             3
Waveform Divisions:             Phrase
Waveform:                       Waveform
Beat Jump Beat Value:           16 Beat
Channel Fader Curve:            Linear
Crossfader Curve:               Fast Cut
Headphones Pre Eq:              Post EQ
Headphones Mono Split:          Stereo
Beat FX Quantize:               On
Mic Low Cut:                    On
Talk Over Mode:                 Advanced
Talk Over Level:                -18dB
MIDI Channel:                   1
MIDI Button Type:               Toggle
Display Brightness:             5
Indicator Brightness:           3
Channel Fader Curve Long Fader: Exponential
Overview Waveform Type:         Half Waveform
Waveform Color:                 Blue
Key Display Format:             Classic
Waveform Current Position:      Center
```

Based on #90 and #91.

### DIFF
--- a/src/device.rs
+++ b/src/device.rs
@@ -128,28 +128,26 @@ impl DeviceExport {
     pub fn get_playlists(&self) -> crate::Result<Vec<PlaylistNode>> {
         match &self.pdb {
             Some(pdb) => pdb.get_playlists(),
-            None => Ok(Vec::new()),
+            None => Err(crate::Error::NotLoadedError),
         }
     }
 
     /// Get the entries for a single playlist.
-    #[must_use]
     pub fn get_playlist_entries(
         &self,
         id: PlaylistTreeNodeId,
-    ) -> Box<dyn Iterator<Item = crate::Result<(u32, TrackId)>> + '_> {
+    ) -> crate::Result<impl Iterator<Item = crate::Result<(u32, TrackId)>> + '_> {
         match &self.pdb {
-            Some(pdb) => Box::new(pdb.get_playlist_entries(id)),
-            None => Box::new(std::iter::empty()),
+            Some(pdb) => Ok(pdb.get_playlist_entries(id)),
+            None => Err(crate::Error::NotLoadedError),
         }
     }
 
     /// Get the tracks.
-    #[must_use]
-    pub fn get_tracks(&self) -> Box<dyn Iterator<Item = crate::Result<Track>> + '_> {
+    pub fn get_tracks(&self) -> crate::Result<impl Iterator<Item = crate::Result<Track>> + '_> {
         match &self.pdb {
-            Some(pdb) => Box::new(pdb.get_tracks()),
-            None => Box::new(std::iter::empty()),
+            Some(pdb) => Ok(pdb.get_tracks()),
+            None => Err(crate::Error::NotLoadedError),
         }
     }
 }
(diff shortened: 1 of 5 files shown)

### CODE REVIEW

File: src/lib.rs [hunk 1]
```diff
@@ -24,10 +24,12 @@
 #![cfg_attr(not(debug_assertions), deny(clippy::used_underscore_binding))]
 
 pub mod anlz;
+pub mod device;
 pub mod pdb;
 pub mod setting;
 pub mod util;
 pub(crate) mod xor;
```
Swiftb0y, line 31: Do we want to make the more internal modules private when they have a nicer API counterpart.

src/device.rs
```diff
@@ -0,0 +1,237 @@
...
+            }
+        });
+
+        settings
+    }
+}
+
+/// Settings object containing for all device settings.
+#[derive(Debug, Default, PartialEq, Eq, Clone)]
+pub struct Settings {
```
Swiftb0y, line 77: What do we want the `Default` to be? Just a struct with all `None`s (which is derived) or with the actual default settings from rekordbox?

src/device.rs
```diff
@@ -0,0 +1,237 @@
...
+    pub waveform_color: Option<setting::WaveformColor>,
+    /// "Key display format" setting.
+    pub key_display_format: Option<setting::KeyDisplayFormat>,
+    /// "Waveform Current Position" setting.
+    pub waveform_current_position: Option<setting::WaveformCurrentPosition>,
+}
+
+impl Settings {
+    fn set_mysetting(&mut self, data: &setting::MySetting) {
+        self.on_air_display = Some(data.on_air_display);
```
Swiftb0y, line 181: fyi this would also work, though the current manual construction is also fine imo.
 ```suggestion
        self.on_air_display = data.on_air_display.into();
```

File: src/main.rs [hunk 4]
```diff
@@ -109,17 +115,354 @@ fn list_playlists(path: &PathBuf) {
...
 
-fn dump_anlz(path: &PathBuf) {
-    let mut reader = std::fs::File::open(&path).expect("failed to open file");
-    let anlz = ANLZ::read(&mut reader).expect("failed to parse setting file");
+fn list_settings(path: &Path) -> rekordcrate::Result<()> {
+    use rekordcrate::DeviceExport;
+
+    let mut export = DeviceExport::default();
+    export.load(path)?;
+    let settings = export.get_settings();
```
Swiftb0y, line 127: IMO constructing a mutable default and then loading from a file doesn't seem very rusty to me. Why not make load a constructor? Also allows for nice chaining.
```suggestion
    let settings = DeviceExport::load(path)?.get_settings();
```

Swiftb0y (commented):
Can you elaborate a bit more on the high-level principle of these Settings? What data is contained in which file, what files are exported, which are expected to be present on an external drive, etc?

Also in terms of code, I'm getting a bit confused because I regularly mix up `Settings` (high-level API) and `Setting` (binrw).

src/lib.rs
Holzhaus: Not sure. Maybe we should group the low-level stuff in a separate submodule or crate but leave it public for people who want to experiment or squeeze out maximum performance?

MarcelRaschke (approved):

src/device.rs
Holzhaus: Rekordbox' default settings probably make sense.

acrilique:
Closing as this was superseded by #210

PR #94, opened 2022-10-13
Column table

Swiftb0y:
~~based on #93 atm~~

most fields have not been reverse-engineered. I just whipped this up while I'm waiting for my windows VM to download so I can create more databases to analyze. From all the databases I've collected so far, the tables were identical, they seem to just contain the defaults I guess.

### DIFF
--- a/src/pdb/mod.rs
+++ b/src/pdb/mod.rs
@@ -76,6 +76,9 @@ pub enum PageType {
     /// Holds rows pointing to album artwork images.
     #[brw(magic = 13u32)]
     Artwork,
+    /// Contains the metadata categories by which Tracks can be browsed by.
+    #[brw(magic = 16u32)]
+    Columns,
     /// Holds information used by rekordbox to synchronize history playlists (not yet studied).
     #[brw(magic = 19u32)]
     History,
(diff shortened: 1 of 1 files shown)


Swiftb0y:
While not fully reverse-engineered, this is in a mergeable state now IMO.

src/pdb/mod.rs
```diff
@@ -664,6 +667,30 @@ pub struct PlaylistEntry {
...
     playlist_id: PlaylistTreeNodeId,
 }
+
+/// Contains the kinds of Metadata Categories tracks can be browsed by
+/// on CDJs.
+#[binrw]
+#[derive(Debug, PartialEq, Eq, Clone)]
+#[brw(little)]
+pub struct ColumnEntry {
+    // possibly the primary key, though I don't know if that would
```
Holzhaus, line 677:
```suggestion
    // Possibly the primary key, though I don't know if that would
```

src/pdb/mod.rs
```diff
@@ -664,6 +667,30 @@ pub struct PlaylistEntry {
...
+    // make sense as I don't think there are references to these
+    // rows anywhere else. This could be a stable ID to identify
+    // a category by in hardware (instead of by name).
+    id: u16,
+    // maybe a bitfield containing infos on sort order and which
```
Holzhaus, line 682:
```suggestion
    // Maybe a bitfield containing infos on sort order and which
```

src/pdb/mod.rs
```diff
@@ -664,6 +667,30 @@ pub struct PlaylistEntry {
...
+    // columns are displayed.
+    unknown0: u16,
+    /// TODO Contained string is prefixed by the "interlinear annotation"
```
Holzhaus, line 685: What exactly is TODO here?

src/pdb/mod.rs
```diff
@@ -948,6 +975,9 @@ pub enum Row {
     /// Represents a track entry in a playlist.
     #[br(pre_assert(page_type == PageType::PlaylistEntries))]
     PlaylistEntry(PlaylistEntry),
+    /// TODO
```
Holzhaus, line 978: Please add a short description here

Holzhaus (commented):
Thanks!

src/pdb/mod.rs
Swiftb0y: Well we need to discuss how we handle these superfluous unicode characters.

src/pdb/mod.rs
Swiftb0y: Sorry, Oversight.

Swiftb0y:
fyi, I got someone else with a CDJ to create a test database and confirm the settings actually work on the CDJ. It seems like the Category and Sort options are not stored in the Columns table for some reason but somewhere else non-obvious. I suspect its either some other of the unknown tables or the `exportExt.pdb`. I don't think its stored in the settings files either.

src/pdb/mod.rs
Holzhaus: I'd say we either keep it like you implemented, or make these separate fields before and after the string, for example`#[br(temp, assert(column_name_before = 0xFFFA)]`.

But the current comment explains how the column name is encoded, but not what's left "to do" here.

```suggestion
    /// Contained string is prefixed by the "interlinear annotation"
```

Swiftb0y: Well you can't really make it a separate field that contains these characters the way DeviceSQLStrings are implemented. I'd rather go with a `#[bw(try_map = ...)]` that maps from a simple `String` (or the higher-level enum) to the DeviceSQLString with the prefixes (and does asserting and stripping when doing the reading). 

IMO with the current implementation, the API is not self-explanatory (since you need to manually add the prefixes when constructing the string), so thats why I think a TODO is appropriate here so we remember to improve that. I'd be happy to reword the TODO if you think we should make this clearer.

Holzhaus (approved):
Thanks

PR #96, opened 2022-10-14
style(lint): only warn on missing docs

Swiftb0y:
Otherwise the build fails when you just want to prototype. The workaround of spamming `/// TODO` everywhere leads to oversights.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -9,11 +9,11 @@
 //! This library provides access to device libraries exported from Pioneer's Rekordbox DJ software.
 
 #![warn(unsafe_code)]
+#![warn(missing_docs)]
 #![cfg_attr(not(debug_assertions), deny(warnings))]
 #![deny(rust_2018_idioms)]
 #![deny(rust_2021_compatibility)]
 #![deny(missing_debug_implementations)]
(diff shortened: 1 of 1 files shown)


Holzhaus:
When prototyping you can just put

    #![allow(missing_docs)]

at the top of the module that you're working on.

Holzhaus:
If we merged this, does pre commit still fail on missing docs?

Swiftb0y:
probably not, maybe we could change the lint level to only `deny(missing_docs)` when clippy is run? Though if that doesn't work reliably just adding `#![allow(missing_docs)]` at the top is probably better. I don't know when I'll have the time to investigate.

Holzhaus:
Just tested this, pre-commit still fails so I see no issue with merging this.

Swiftb0y:
Thank you. Glad to see you finding the time to revive this repo. I'll probably don't have the time to do reverse-engineering own my own right now but I'll try to help out in the form of reviews as much as I can.
IIRC there has been some new insight being shared on https://deep-symmetry.zulipchat.com/#narrow/stream/275855-dysentery-.26-crate-digger which allowed some people to create mostly working exports.

PR #98, opened 2022-11-05
refactor(pdb): Use `bitflags` crate for page flags

### LINKED ISSUE #95: Improve `pdb::PageFlags` struct
We already have a dependency on the [`modular_bitfield` crate](https://docs.rs/modular-bitfield/latest/modular_bitfield/), might make sense to use that instead:

```rust
#[bitfield]
pub struct PageFlags {
    unknown1: B1,
    no_data: bool,
    unknown2: B6,
}
```

_Originally posted by @Holzhaus in https://github.com/Holzhaus/rekordcrate/pull/93#discussion_r993200894_

Holzhaus:
Idea how to improve the pageflags struct. Unsure if it already makes sense to merge this, because it introduces an additional dependency.

Resolves #95.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -275,6 +275,7 @@ name = "rekordcrate"
 version = "0.2.0"
 dependencies = [
  "binrw",
+ "bitflags",
  "clap",
  "crc16",
  "glob",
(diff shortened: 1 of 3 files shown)


Swiftb0y:
Do we want yet another convenience proc macro crate as a dependency?

Holzhaus:
That's why it's a draft. Wasn't sure about this. If we expose that flags type (and have more than one), it might make sense. Right now, it doesn't really.

PR #101, opened 2023-11-14
update binrw to 0.13

erictapen:
A few changes accumulated upstream, so I thiought it might be worth to migrate.

I'm not super savvy around lifetimes in Rust, so these might benefit from a review. But the tests run through.


src/util.rs
```diff
@@ -81,44 +81,45 @@ pub(crate) mod testing {
...
         bin: &[u8],
         obj: T,
-        read_args: <T as binrw::BinRead>::Args,
-        write_args: <T as binrw::BinWrite>::Args,
+        read_args: <T as binrw::BinRead>::Args<'a>,
+        write_args: <T as binrw::BinWrite>::Args<'a>,
     ) where
+        <T as binrw::BinRead>::Args<'a>: Clone,
+        <T as binrw::BinWrite>::Args<'a>: Clone,
         T: BinRead + BinWrite + PartialEq + core::fmt::Debug + ReadEndian + WriteEndian,
```
Swiftb0y, line 92: This looks a little weird to me, if we require the `Args` to satisfy `Clone` anyways, can we avoid `'a` on the function somehow?

Swiftb0y (changes requested):
Hey there, glad to see this library receiving some love. I have a couple questions in regards to your changes.
My rust has also gotten quite rusty (in the bad way ;) ), so I apologize if my review comments are dumb.

I think we need to look a little closer at `FilePtr` in order to upgrade to binrw 0.12. since some semantics for it have changed and we need to ensure that didn't break anything. If the unit tests succeed, it should work, but I need @Holzhaus to approve those to run.

Holzhaus:
Thanks for the contribution. I approved the check, let's see what CI says.

src/pdb/mod.rs
```diff
@@ -408,14 +409,14 @@ impl RowGroup {
     const MAX_ROW_COUNT: u16 = 16;
 
     /// Return the ordered list of row offsets that are actually present.
-    pub fn present_rows(&self) -> impl Iterator<Item = &Row> {
+    pub fn present_rows(&self) -> impl Iterator<Item = Row> + '_ {
```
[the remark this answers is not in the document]
erictapen: Yeah I was wondering about that change too. With the old behaviour it yields the following error.
```
error[E0271]: expected `[closure@mod.rs:417:25]` to be a closure that returns `Option<&Row>`, but it returns `Option<Row>`
   --> src/pdb/mod.rs:412:35
    |
412 |     pub fn present_rows(&self) -> impl Iterator<Item = &Row> {
    |                                   ^^^^^^^^^^^^^^^^^^^^^^^^^^ expected `Option<Row>`, found `Option<&Row>`
    |
    = note: expected enum `Option<Row>`
               found enum `Option<&Row>`
    = note: required for `FilterMap<Enumerate<Rev<Iter<'_, FilePtr<u16, Row>>>>, ...>` to implement `Iterator`
    = note: the full type name has been written to '/home/kerstin/git/rekordcrate/target/debug/deps/rekordcrate-b18dc1c7427bce73.long-type-18347493313743629262.txt'
```

Maybe this is due to binrw changes to `FilePtr`? I couldn't grasp a relevant item in their release notes though.

src/util.rs
```diff
@@ -81,44 +81,45 @@ pub(crate) mod testing {
...
 
-    pub fn test_roundtrip<T>(bin: &[u8], obj: T)
+    pub fn test_roundtrip<'a, T>(bin: &[u8], obj: T)
     where
-        <T as binrw::BinRead>::Args: Default,
-        <T as binrw::BinWrite>::Args: Default,
-        T: BinRead + BinWrite + PartialEq + core::fmt::Debug + ReadEndian + WriteEndian,
+        <T as binrw::BinRead>::Args<'a>: Default + Clone,
+        <T as binrw::BinWrite>::Args<'a>: Default + Clone,
+        T: BinRead + BinWrite + PartialEq + core::fmt::Debug + ReadEndian + WriteEndian + Clone,
```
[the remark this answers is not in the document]
erictapen: True! I removed it.

src/util.rs
erictapen: Mh, I'm unsure how to achieve that. When just removing the `'a` from the function signature, it yields this:

```
error[E0261]: use of undeclared lifetime name `'a`
  --> src/util.rs:90:37
   |
90 |         <T as binrw::BinRead>::Args<'a>: Clone, 
   |                                     ^^ undeclared lifetime
   |
   = note: for more information on higher-ranked polymorphism, visit https://doc.rust-lang.org/nomicon/hr
tb.html
help: consider making the bound lifetime-generic with a new `'a` lifetime
   |
90 |         for<'a> <T as binrw::BinRead>::Args<'a>: Clone,
   |         +++++++
help: consider introducing lifetime `'a` here
   |
84 |     pub fn test_roundtrip_with_args<'a, T>(
   |                                     +++


```

Setting the lifetime seems to be required for `Args` and leaving it empty as in `'_` doesn't work either.

Swiftb0y: Yeah, thats weird. I'm not deep enough into Rust to understand this either. I guess we can ignore it until someone shows up with (or I acquire it) the skill to look into this.

src/pdb/mod.rs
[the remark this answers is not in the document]
Swiftb0y: Okay I got a little farther this now. So yeah, changes in `FilePtr` are indirectly to blame. From the 0.12 changelog: "The default behaviour of FilePtr has been changed to always immediately seek to and read the pointed-to value." This implies that FilePtr.value is not an optional anymore and thus `as_ref` is not implemented (or something along the lines). The signature can be kept the same, we just need to return a `Some(&row_offset.value)` instead. Now I'm just facing some ownership issues in the callers but those don't seem to be hard to resolve. I'll file a PR to your fork once I have found a solution.

Holzhaus:
Sorry, I'm very short on time atm. @SWIFTB0Y Ready to merge from your POV?

Swiftb0y:
Same here unfortunately. I'm not super happy about the forced clone but we can worry about that later, it makes more sense to keep the dependencies up-to-date.

src/pdb/mod.rs
```diff
@@ -408,14 +409,14 @@ impl RowGroup {
...
         self.rows
             .iter()
             .rev()
             .enumerate()
             .filter_map(|(i, row_offset)| {
                 if (self.row_presence_flags & (1 << i)) != 0 {
-                    row_offset.value.as_ref()
+                    Some(row_offset.value.clone())
```
Swiftb0y, line 419:
```suggestion
                    // TODO: the explicit clone is probably quite expensive
                    // but the simplest way to make the borrow checker happy for now. 
                    // This is forced by the changes to FilePtr in binrw 0.12.
                    // We should investigate how to remove the clone in the future.
                    Some(row_offset.value.clone())
```

Swiftb0y (approved):
Please add a comment to highlight the clone (feel free to use my suggestion as is). LGTM otherwise. Thank you very much.

erictapen:
Done!

Swiftb0y:
@Holzhaus this PR is ready. Would you be so kind and approve the workflows and merge the PR afterwards? I don't have the rights to do either. Thank you.

Holzhaus:
@erictapen Thank you very much for your contribution and and also thank you @Swiftb0y for your review.

There was a code style issue due to trailing whitespace in a comment. I resolved this and pushed directly to the fork. I'll merge now.

erictapen:
Thanks a lot for review and merging! @summersamara and I worked on pdb serialization a bit, will open our next PR in the coming days.

PR #103, opened 2024-02-26
PDB Serialization Pt. 4, add reexport-pdb command

erictapen:
@summersamara and I worked a bit on the issue of PDB serialization. Our efforts are far from being mergeable, but we managed to generate PDB files containing tracks, that can be read by Rekordbox software.

This builds on top of https://github.com/Holzhaus/rekordcrate/pull/100.

For easier testing, we introduce the `reexport-pdb` command, that deserializes an existing PDB file and then serializes the internal representation again. Coming from Rekordbox generated databases this never results in indentical outputs, but we observed the result to be readable by Rekordbox. Also reexporting an already reexported file seems to return an identical file.

Originally we wanted to bring our commits in a shape compliant with CONTRIBUTING.md, but never got to it and now I thought it might be better to at least share our efforts for future experimentation than to let our progress bitrot on disk.

Looking forward to your feedback.


Holzhaus:
Awesome, thanks. I'm currently very busy with work, but I'll have a closer look soon-ish. 

Also pinging @Swiftb0y :)

Swiftb0y:
Same thing here. I very much appreciate your PR, but I think we should look at #100 first. It would really help if you @erictapen could also do an initial review on that to see if you spot any obvious issues.

erictapen:
@Swiftb0y Sounds sensible, but #100 contains conflicts, as it lacks the patches for binrw 0.13. Not sure how easy it is to pluck them from this PR.

Swiftb0y:
Good point. @Holzhaus do you have time to bring #100 into shape again?

PR #110, opened 2024-08-09
Fix missing table rows

### LINKED ISSUE #107: Missing rows table rows.
As per https://github.com/Deep-Symmetry/crate-digger/issues/32

https://github.com/Holzhaus/rekordcrate/blob/01ebd87da36e19488d17f1fc0248b6e6802978ea/src/pdb/mod.rs#L324-L350

Needs fixing

Holzhaus:
Fixes #107.

Also see https://github.com/Deep-Symmetry/crate-digger/issues/32.


Holzhaus:
@Swiftb0y I would appreciate your review :)

src/pdb/mod.rs
```diff
@@ -313,37 +300,41 @@ impl Page {
...
+            reader.seek(
+                u64::try_from(row_groups.len())
+                    .ok()
+                    .and_then(|index| index.checked_mul(36))
+                    .and_then(|x| stream_position.checked_sub(x))
+                    .map(SeekFrom::Start)
+                    .ok_or_else(|| binrw::Error::AssertFail {
+                        pos: stream_position,
+                        message: format!("Failed to calculate seek position for row group {}", i),
+                    })?,
```
Swiftb0y, line 335: Maybe I'm a little narrow minded from too much C++ programming but this feels overcautious. `row_groups.len()` is a `usize` and will pretty much always fit into a `u64`, right? Are there architectures where usize is bigger than a u64? can we `static_assert` that (is there such a thing in `rust`)?

Swiftb0y (commented):
a couple mostly superficial comments. I have not dug deeply into https://github.com/Deep-Symmetry/crate-digger/issues/32 so I can't really assert how well this implementation aligns with the spec.

Very happy to see some activity here btw.

brunchboy:
The questions Jan Holthuis raised in the Crate Digger issue make me concerned that we are not calculating the number of row groups correctly, and perhaps we simply can’t calculate that in advance at all, because we don’t know how many rows have been invalidated in each row group, so we may simply need to scan row groups until we have seen `num_rows` non-deleted rows. But it sounded like he found a database where there were simply not that many non-deleted rows in the page at all. Perhaps it was corrupt?

Swiftb0y:
fyi, I can't resolve my review comments, so take the entire review as addressed after merging #111

PR #111, opened 2024-08-14
Apply review comments to #110

Swiftb0y:
see commit messages for explanations (feel free to squash into original commit after merging if you prefer that). While trying to apply my review suggestions, I found some of them to not result in improvements, so see those are resolved.

### DIFF
--- a/src/pdb/mod.rs
+++ b/src/pdb/mod.rs
@@ -394,11 +394,11 @@ impl BinRead for RowGroup {
         endian: Endian,
         (page_type, page_heap_offset): Self::Args<'_>,
     ) -> BinResult<Self> {
-        let stream_position = reader.stream_position()?;
+        let table_page_end_position = reader.stream_position()?;
         reader.seek(SeekFrom::Current(-4))?;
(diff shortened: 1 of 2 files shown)


src/pdb/mod.rs
```diff
@@ -400,11 +394,11 @@ impl BinRead for RowGroup {
         endian: Endian,
         (page_type, page_heap_offset): Self::Args<'_>,
     ) -> BinResult<Self> {
-        let stream_position = reader.stream_position()?;
+        let table_page_end_position = reader.stream_position()?;
```
Holzhaus, line 397: The new name is more descriptive, but unfortunately wrong if there is more than one row group. It's the end position of the current row group.

Maybe 
```suggestion
        let row_group_end_position = reader.stream_position()?;
```

Swiftb0y: whoops, yeah you're right. Fixup'd

Swiftb0y:
do you want me to apply the fixup now or whats the next step?

Swiftb0y:
No worries, thank you very much.

Swiftb0y:
Hope you had a nice vacation btw ;)

Issue #127, opened 2025-01-21
$200 bounty for Rekordbox export with hot cues & BPM


HarryR:
The [XML format][1], despite being "legacy", is valuable precisely because:

 * It's an open, human-readable format that can be easily generated and used by other tools
 * It has widespread compatibility across different hardware & software (NI/Traktor, Virtual DJ, Serato, Denon)
 * Most importantly - you can build your own tools around it without dealing with Pioneer's proprietary formats or depending on Rekordbox's stability

Mostly I'm concerned about having the first 3 or 4 hot cues, and BPM, for each track in the playlist (with artist & title), as it's the lowest common denominator for what's supported even on ancient gear.

[1]: https://cdn.rekordbox.com/files/20200410160904/xml_format_list.pdf

However, the bounty is available for any video that shows exporting a playlist from Mixxx, plugging it into a mid-2010s Pioneer CDJ (basically 10 year old equipment), skipping to a hot cue and syncing the other deck to the same BPM etc.

I know this is more complicated than it sounds, as even RekordBox has problems with accidentally overwriting everything.
 
Payable in USDC or equivalent crypto preferable.

Holzhaus:
I still have some code for XML support around that I never finished. I'll look into it when I find the time.

Holzhaus:
@HarryR do you have some real-world XMLs that you can provide? Then I can add them as test cases for robustness.

HarryR:
> [@HarryR](https://github.com/HarryR) do you have some real-world XMLs that you can provide? Then I can add them as test cases for robustness.

Unfortunately I don't have access to my friends equipment at the moment, I'll ask.

In the meantime may be worth posting on Reddit or the Mixxx forums?

acrilique:
Seems like anything with the word "bounty" will unavoidably attract bots...

PR #129, opened 2025-01-23
feat(xml): Add support for Rekordbox XML format

Holzhaus:
Adds the XML format support requested in #127.


src/main.rs
```diff
@@ -159,6 +166,15 @@ fn dump_setting(path: &PathBuf) -> rekordcrate::Result<()> {
     Ok(())
 }
 
+fn dump_xml(path: &PathBuf) -> rekordcrate::Result<()> {
+    let file = std::fs::File::open(path)?;
+    let reader = std::io::BufReader::new(file);
+    let document: Document = quick_xml::de::from_reader(reader).expect("failed to deserialize XML");
```
Swiftb0y, line 172: does this need to panic? Would've something like this make more sense?
```suggestion
    let document: Document = quick_xml::de::from_reader(reader).map_err(|_| "failed to deserialize XML")?;
```

Swiftb0y (commented):
small nitpick. `src/xml.rs` is above me unfortunately so I can't comment much there. As a followup, it would be nice if the XML was a crate feature, so consumers that don't need it can avoid pulling in serde.

src/main.rs
Holzhaus: Yeah makes sense.

PR #130, opened 2025-02-02
XML Improvements

Holzhaus:


src/main.rs
```diff
@@ -166,10 +166,12 @@ fn dump_setting(path: &PathBuf) -> rekordcrate::Result<()> {
     Ok(())
 }
 
+#[cfg(feature = "xml")]
 fn dump_xml(path: &PathBuf) -> rekordcrate::Result<()> {
     let file = std::fs::File::open(path)?;
     let reader = std::io::BufReader::new(file);
-    let document: Document = quick_xml::de::from_reader(reader).expect("failed to deserialize XML");
+    let document: rekordcrate::xml::Document =
+        quick_xml::de::from_reader(reader).expect("failed to deserialize XML");
```
Swiftb0y, line 174: have you figured out how to avoid the `expect` here yet? (the thing I criticized in the previous PR).

src/xml.rs
```diff
@@ -27,221 +33,429 @@ pub struct Document {
...
+            tracks: &'a Vec<Track>,
+        }
+
+        let value = Value {
+            entries: self.tracks.len(),
+            tracks: &self.tracks,
+        };
+
+        value.serialize(serializer)
+    }
```
Swiftb0y, line 95: I feel like this could be simplified by using `#[serde(getter = "tracks.len()")]` instead?

src/xml.rs
```diff
@@ -27,221 +33,429 @@ pub struct Document {
...
+impl Serialize for Collection {
+    fn serialize<S>(&self, serializer: S) -> Result<S::Ok, S::Error>
+    where
+        S: Serializer,
+    {
+        #[derive(Serialize)]
+        struct Value<'a> {
+            /// Number of TRACK in COLLECTION
+            #[serde(rename = "@Count")]
+            entries: usize,
```
Swiftb0y, line 83: The commit message doesn't explain why this is now `@Count` when it was previously `@Entries`. Can you elaborate?

src/xml.rs
```diff
@@ -27,221 +33,429 @@ pub struct Document {
...
+        match self {
+            Self::Rose => write!(f, "Rose"),
+            Self::Red => write!(f, "Red"),
+            Self::Orange => write!(f, "Orange"),
+            Self::Lemon => write!(f, "Lemon"),
+            Self::Green => write!(f, "Green"),
+            Self::Turquoise => write!(f, "Turquoise"),
+            Self::Blue => write!(f, "Blue"),
+            Self::Violet => write!(f, "Violet"),
+            Self::Custom(rgb) => write!(f, "#{:02X}{:02X}{:02X}", rgb.r, rgb.g, rgb.b),
```
Swiftb0y, line 336: Alternative: delegate to the inner type if you don't need a stable representation 
https://docs.rs/rgb/latest/rgb/type.RGB8.html#impl-Display-for-Rgb%3CT%3E
```suggestion
            Self::Custom(rgb) => write!(f, "{}", rgb),
```

src/xml.rs
```diff
@@ -27,221 +33,429 @@ pub struct Document {
...
+    One = 0x33,
+    /// Two Stars
+    Two = 0x66,
+    /// Three Stars
+    Three = 0x99,
+    /// Four Stars
+    Four = 0xCC,
+    /// Five Stars
+    Five = 0xFF,
 }
```
Swiftb0y, line 393: since this is a `u8` under the hood, do fractional stars also exist (eg `230` corresponding to roughly `4.5` stars)? Do those need to be modeled here?

Swiftb0y (commented):
large lgtm, couple thoughts

src/xml.rs
Holzhaus: The format is inconsistent, and it think I confused the item count for the Collection element ("Entries") with the item count for the Node element ("Count"). The original version was correct. I will revert.

image

src/xml.rs
Holzhaus: I think Rekordbox UI doesn't allow this. We could of course try to make a `Custom(u8)` variant, but I don't think it's needed. If it turns out to be necessary, we'll bump the version :)

acrilique (approved):
This LGTM, may I take the freedom to fix the conflicts and merge?

acrilique:
Just rebased fixing all conflicts and added a fix and a test for #145 , thus superseding #146
I'll leave this here for a few days to give you all the change to review.

acrilique:
Merging this to keep moving forward despite the lack of feedback

PR #135, opened 2025-02-24
CI Hardening

Holzhaus:


.github/workflows/docs.yml
```diff
@@ -1,59 +1,67 @@
...
+                      f.write(doc)
+              env:
+                  PACKAGE_NAME: rekordcrate
+                  OUTPUT_DIR: target/doc
+
+            - name: Deploy to GitHub Pages
+              uses: peaceiris/actions-gh-pages@v4
+              with:
+                  github_token: ${{ secrets.GITHUB_TOKEN }}
+                  publish_dir: target/doc
```
Swiftb0y, line 67: Why do you do this if the docs are also on docs.rs?

Holzhaus: For in-development docs that are not yet tagged. But I'm not passionate about this, we could also remove it.

Swiftb0y: nah its fine. Just didn't know we had this. For local dev I feel like `cargo doc --open` suffices too though.

Swiftb0y (approved):

Issue #145, opened 2025-04-01
xml parsing fails if a playlist contains no tracks


lazulit3:
If your exported XML includes a rekordbox playlist that contains no tracks, XML deserialization will fail.

For example, if you add an empty playlist to `data/xml/database.xml` like:

```diff
diff --git a/data/xml/database.xml b/data/xml/database.xml
index dd29426..8186f3a 100644
--- a/data/xml/database.xml
+++ b/data/xml/database.xml
@@ -61,7 +61,7 @@
     </TRACK>
   </COLLECTION>
   <PLAYLISTS>
-    <NODE Type="0" Name="ROOT" Count="2">
+    <NODE Type="0" Name="ROOT" Count="3">
       <NODE Name="Folder" Type="0" Count="1">
         <NODE Name="Sub Playlist" Type="1" KeyType="0" Entries="2">
           <TRACK Key="5"/>
@@ -72,6 +72,7 @@
         <TRACK Key="5"/>
         <TRACK Key="6"/>
       </NODE>
+      <NODE Name="EmptyPlaylist" Type="1" KeyType="0" Entries="0" />
     </NODE>
   </PLAYLISTS>
 </DJ_PLAYLISTS>
```

Then try parsing the XML:

```sh
cargo run dump-xml data/xml/database.xml
```

This produces an error:
```
thread 'main' panicked at src/main.rs:172:65:
failed to deserialize XML: Custom("missing field `TRACK`")
```

I have not tested this with  an empty folder, but I suspect the same bug exists if your Rekordbox collection includes empty folders as well.

PR #146, opened 2025-04-01
fix(xml): XML deserialization to handle empty nodes 

lazulit3:
This PR fixes a bug where XML parsing would panic if the exported XML contains an edge case where there is (either of):
- a folder with no children
- a playlist with no tracks

The bug is described at:
- https://github.com/Holzhaus/rekordcrate/issues/145

## Approach

This removes the visitor map implementation. Instead, a node is first deserialized into a general `Node` type (which can contain fields for either a folder or playlist). After identifying which type the node is, the specific type is constructed and returned.

## Discussion

I am not sure if this approach is advantageous, but I struggled to wrap my head around the visitor map.

This PR is against `main`, but I can open a PR against `xml-improvements` as well (if desired).

## Testing

I added a test case for an empty playlist, and this seems to resolve the issue. Additionally, this branch parses my rekordbox export (with my real data) successfully.

### DIFF
--- a/src/xml.rs
+++ b/src/xml.rs
@@ -17,7 +17,6 @@
 //! - <https://pyrekordbox.readthedocs.io/en/stable/formats/xml.html>
 type NaiveDate = String; //Replace with "use chrono::naive::NaiveDate;"
 use serde::{de::Error, ser::Serializer, Deserialize, Serialize};
-use std::borrow::Cow;
 
 /// The XML root element of a rekordbox XML file.
 #[derive(Debug, PartialEq, Clone, Serialize, Deserialize)]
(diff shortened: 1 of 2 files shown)


src/xml.rs
```diff
@@ -251,110 +250,47 @@ impl<'de> Deserialize<'de> for PlaylistGenericNode {
...
-                                    // Create anonymous type
-                                    #[derive(serde::Deserialize)]
-                                    struct Nodes {
-                                        #[serde(rename = "NODE")]
-                                        content: Vec<PlaylistGenericNode>,
-                                    }
-                                    let de = serde::de::value::MapAccessDeserializer::new(map);
-                                    Nodes::deserialize(de)?.content
-                                };
-                                // FIXME: Should we check if nodes.len() == count here?
```
lazulit3, line 301: Should I restore the `FIXME` comments regarding validation that the amount of parsed children matches `Entries` or `Count`?

I could also add this validation, but I was unsure what type of error to return (or would we rather panic?)

src/xml.rs
```diff
@@ -251,110 +250,47 @@ impl<'de> Deserialize<'de> for PlaylistGenericNode {
...
+            // appears on playlists only
+            #[serde(rename = "@KeyType", default)]
+            key_type: Option<String>,
+            // child nodes in a folder
+            #[serde(rename = "NODE", default)]
+            nodes: Vec<PlaylistGenericNode>,
+            // tracks in a playlist
+            #[serde(rename = "TRACK", default)]
+            tracks: Vec<PlaylistTrack>,
+        }
```
lazulit3, line 270: Note that this does not deserialize `Count` or `Entries` because this iteration does not actually use these.

src/xml.rs
```diff
@@ -251,110 +250,47 @@ impl<'de> Deserialize<'de> for PlaylistGenericNode {
...
+            // Playlist node
+            "1" => {
+                if let Some(key_type) = node.key_type {
+                    Ok(PlaylistGenericNode::Playlist(PlaylistPlaylistNode {
+                        name: node.name,
+                        keytype: key_type,
+                        tracks: node.tracks,
+                    }))
+                } else {
+                    Err(D::Error::missing_field("@KeyType"))
```
lazulit3, line 289: This does not actually validate that the `Entries` or `Count` attributes exist on the nodes. Would we prefer this to be stricter?

lazulit3 (commented):
Some notes & discussion included below.

src/xml.rs
Swiftb0y: I'm unfortunately very unfamiliar with the XML code, so I would prefer to stick as close as possible to the previous implementation (which was quite strict from what I can tell).

src/xml.rs
Swiftb0y: Yeah please add it if feasible. Maybe just add a (non-fatal) warning though, as we don't control the XML schema, we want to be maximally compatible

Swiftb0y (commented):
hey there, I'm sorry for the long delay. Are you still interested in merging this PR?

acrilique:
superseded by #130

PR #162, opened 2025-08-31
fix(pdb): get track_page test closer to passing

acrilique:
Still needs some work to properly calculate the offsets at the end of the page, in order to pass the test.
[#development > export to rekordbox USB format @ 💬](https://mixxx.zulipchat.com/#narrow/channel/109171-development/topic/export.20to.20rekordbox.20USB.20format/near/536974521)

### DIFF
--- a/src/pdb/mod.rs
+++ b/src/pdb/mod.rs
@@ -204,12 +204,16 @@ impl PageFlags {
 /// offset found in the page footer at the end of the page.
 #[binread]
 #[derive(Debug, PartialEq)]
-#[br(little, magic = 0u32)]
+#[br(little)]
 #[br(import(page_size: u32))]
 pub struct Page {
-    /// Stream position immediately after magic; used to compute heap base for standalone buffers.
+    /// Stream position at the beginning of the page; used to compute heap base for standalone buffers.
     #[br(temp, parse_with = current_offset)]
-    page_pos_after_magic: u64,
+    page_start_pos: u64,
+    /// Magic signature for pages (must be 0).
+    #[br(temp, assert(magic == 0u32))]
+    #[bw(ignore)]
+    magic: u32,
     /// Index of the page.
     ///
     /// Should match the index used for lookup and can be used to verify that the correct page was loaded.
@@ -285,15 +289,13 @@ pub struct Page {
     ///
     /// **Note:** This is a virtual field and not actually read from the file.
     #[br(temp)]
-    // TODO: Use `num_rows.div_ceil(RowGroup::MAX_ROW_COUNT)` here when it becomes available
-    // (currently nightly-only, see https://github.com/rust-lang/rust/issues/88581).
-    #[br(calc = if num_rows > 0 { (num_rows - 1) / RowGroup::MAX_ROW_COUNT + 1 } else { 0 })]
+    #[br(calc = num_rows.div_ceil(RowGroup::MAX_ROW_COUNT))]
     num_row_groups: u16,
     /// The offset at which the row data for this page are located.
     ///
     /// **Note:** This is a virtual field and not actually read from the file.
     #[br(temp)]
-    #[br(calc = (page_pos_after_magic - 4) + u64::from(Self::HEADER_SIZE))]
+    #[br(calc = page_start_pos + u64::from(Self::HEADER_SIZE))]
     page_heap_offset: u64,
     /// Row groups belonging to this page.
     #[br(seek_before(SeekFrom::Current(Page::row_groups_offset(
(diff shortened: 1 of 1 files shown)


Swiftb0y:
Thanks, do you want to look into the next part, why parsing the buffer fails?

Swiftb0y:
Interesting... I mean that somewhat makes sense. The rows in relational databases generally don't have a deterministic order. I guess the reversed order stems from the fact that I reversed the order of the offsets. You can try to remove the `.rev()` here and see if that fixes it or breaks any other test.

https://github.com/Holzhaus/rekordcrate/blob/3dcc01742e3813ce2cab4c59cd4c5af8656a63b2/src/pdb/mod.rs#L417

Swiftb0y:
>   1) The base used to interpret row pointers during parsing of an isolated Page in tests did not match the actual stream position (we were computing offset from page_index × page_size instead of the in-buffer position).

Yeah I suspected something similar.

>   2) The row group footer did not always emit the exact number of offsets the reader expects (and used a filtered scheme with 0xFFFF sentinels). DeviceSQL stores exactly N 16-bit offsets per group (reverse order), and uses 0x0000 when there is no row in a slot.

That was not a problem IMO. The offsets stored there can be arbitrary because the offset pointers are not followed (or at least shouldn't be) if the row is not marked present in the mask following the offsets.

src/pdb/mod.rs
```diff
@@ -207,6 +207,9 @@ impl PageFlags {
 #[br(little, magic = 0u32)]
 #[br(import(page_size: u32))]
 pub struct Page {
+    /// Stream position immediately after magic; used to compute heap base for standalone buffers.
+    #[br(temp, parse_with = current_offset)]
+    page_pos_after_magic: u64,
```
Swiftb0y, line 212: consider reordering here so you first record the temp page_pos and then assert the magic in a separate private field. That would avoid having to mess with questionable index shifts like `- 4`

src/pdb/mod.rs
```diff
@@ -513,74 +540,55 @@ impl RowGroup {
         &self,
         writer: &mut W,
         options: &WriteOptions,
-        args: (u64, u64),
+        args: (u64, u64, u16),
     ) -> binrw::BinResult<u64> {
         let options = options.with_endian(Endian::Little);
 
-        let (page_offset, relative_row_offset) = args;
+        let (page_offset, relative_row_offset, rows_in_group) = args;
```
Swiftb0y, line 547: you can just use `self.rows.len()` instead of passing `rows_in_group` through here.

src/pdb/mod.rs
```diff
@@ -353,11 +356,33 @@ impl BinWrite for Page {
...
+            self.num_rows_large
+        } else {
+            self.num_rows_small.into()
+        };
+
+        let num_groups: u16 = if total_rows > 0 {
+            (total_rows - 1) / RowGroup::MAX_ROW_COUNT + 1
+        } else {
+            0
+        };
```
Swiftb0y, line 371: isn't this just [`u16::div_ceil`](https://doc.rust-lang.org/std/primitive.u16.html#method.div_ceil)? (please double check).

src/pdb/mod.rs
```diff
@@ -353,11 +356,33 @@ impl BinWrite for Page {
...
+        for (idx, row_group) in self.row_groups.iter().enumerate() {
+            let is_last = (idx as u16) == num_groups.saturating_sub(1);
+            let mut rows_in_group = if is_last {
+                total_rows % RowGroup::MAX_ROW_COUNT
+            } else {
+                RowGroup::MAX_ROW_COUNT
+            };
+            if rows_in_group == 0 && total_rows > 0 {
+                rows_in_group = RowGroup::MAX_ROW_COUNT;
+            }
```
Swiftb0y, line 381: I can't wrap my head around this arithmetic. Can you explain it to me?

Swiftb0y (commented):
couple thoughts. Thanks so far.

acrilique:
Thanks a lot for the feedback. I just applied most of your suggestions, which simplified/clarified the code while still passing the test. Please if you can, check if I've missed anything.

src/pdb/mod.rs
```diff
@@ -204,12 +204,16 @@ impl PageFlags {
...
 pub struct Page {
-    /// Stream position immediately after magic; used to compute heap base for standalone buffers.
+    /// Stream position at the beginning of the page; used to compute heap base for standalone buffers.
     #[br(temp, parse_with = current_offset)]
-    page_pos_after_magic: u64,
+    page_start_pos: u64,
+    /// Magic signature for pages (must be 0).
+    #[br(temp, assert(magic == 0u32))]
+    #[bw(ignore)]
+    magic: u32,
```
Swiftb0y, line 216: Note to self: investigate whether we actually write the magic bytes too.

src/pdb/mod.rs
```diff
@@ -285,15 +289,13 @@ pub struct Page {
     ///
     /// **Note:** This is a virtual field and not actually read from the file.
     #[br(temp)]
-    // TODO: Use `num_rows.div_ceil(RowGroup::MAX_ROW_COUNT)` here when it becomes available
-    // (currently nightly-only, see https://github.com/rust-lang/rust/issues/88581).
-    #[br(calc = if num_rows > 0 { (num_rows - 1) / RowGroup::MAX_ROW_COUNT + 1 } else { 0 })]
+    #[br(calc = num_rows.div_ceil(RowGroup::MAX_ROW_COUNT))]
```
Swiftb0y, line 292: thanks ;)

src/pdb/mod.rs
```diff
@@ -540,11 +517,12 @@ impl RowGroup {
...
         writer: &mut W,
         options: &WriteOptions,
-        args: (u64, u64, u16),
+        args: (u64, u64),
     ) -> binrw::BinResult<u64> {
         let options = options.with_endian(Endian::Little);
 
-        let (page_offset, relative_row_offset, rows_in_group) = args;
+        let (page_offset, relative_row_offset) = args;
+        let rows_in_group: u16 = self.rows.len() as u16;
```
Swiftb0y, line 525: NIT: This silently overflows. Use a try_into instead and properly handle the error to the parser. We want this to be maximally robust.

acrilique: Apart from using try_into to avoid overflows should we check `len()` does not exceed `MAX_ROW_COUNT`?

src/pdb/mod.rs
```diff
@@ -1166,7 +1185,9 @@ pub enum Row {
 }
 
 impl Row {
-    #[must_use]
+    // TODO We should survey all the different kind of rows and ensure they're 
+    // aligned properly. Temporally allowed unused
+    #[allow(unused)]
```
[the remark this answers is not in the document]
acrilique: done

Swiftb0y:
this is not a blocker for now, but for future commits, you should install pre-commit. Otherwise it'll get annoying to satisfy pre-commit once the branch is properly rebased and CI runs in full again.

acrilique:
> this is not a blocker for now, but for future commits, you should install pre-commit. Otherwise it'll get annoying to satisfy pre-commit once the branch is properly rebased and CI runs in full again.

Okay, sorry I did not read the contribution guide before so I was not aware of pre-commit. I've just installed it.

Swiftb0y:
As pointed out yesterday, I will merge now to keep velocity high.

PR #169, opened 2025-09-15
feat(pdb): implement index pages (de-)serialization

acrilique:


src/pdb/mod.rs
```diff
@@ -127,9 +128,19 @@ pub enum PlainPageType {
 #[binrw]
 #[derive(Clone, Debug, PartialEq, Eq, PartialOrd)]
 #[brw(little)]
-pub struct PageIndex(u32);
+pub struct PageIndex(pub u32);
```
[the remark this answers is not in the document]
Swiftb0y: yeah, consider removing that, not big stress though.

Swiftb0y (approved):
Thank you. I'll stop nitpicking now. This look good. I'll let this rest for a couple days so @Holzhaus can take a look if he wants to.

src/pdb/mod.rs
```diff
@@ -234,31 +241,205 @@ impl PageFlags {
...
+
+        let content_size = page_size - Page::HEADER_SIZE;
+        let padding_end_offset = content_size - 20;
+
+        /* Fill with empty entries (0x1ffffff8) until the last 20 bytes, which
+        are zeroes. If https://github.com/jam1garner/binrw/issues/205 was ever
+        fixed, this entire BinWrite implementation could possibly be removed.
+        */
+
+        if written_bytes < padding_end_offset as u64 {
```
Holzhaus, line 384: Let's try to avoid `as` if `Into`/`From` also work. Reason is that if the types change so that this conversation is no longer lossless, the compiler will warn us about it. When using `as`, stuff would be silently lost. 

```suggestion
        if written_bytes < u64::from(padding_end_offset) {
```

Holzhaus (commented):
Thanks, I didn't have the time to test yet, but code looks good! Only some small nitpicks.

acrilique:
Thanks a lot for your comments. They should all be addressed by my last commit

src/pdb/mod.rs
```diff
@@ -234,31 +241,205 @@ impl PageFlags {
...
+        self.0 == 0x1FFF_FFF8
+    }
+
+    /// Creates a new `IndexEntry` from a `PageIndex` and `index_flags`.
+    #[must_use]
+    pub const fn new(page_index: PageIndex, index_flags: u8) -> Option<Self> {
+        if index_flags & 0b111 != index_flags {
+            return None;
+        }
+        Some(Self((page_index.0 << 3) | (index_flags & 0b111) as u32))
```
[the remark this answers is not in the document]
Swiftb0y: yeah, unfortunately the functions implemented on bools are not const _anymore_ for some reason. Lets just go with the the manual construction for now.

src/pdb/mod.rs
```diff
@@ -234,31 +258,206 @@ impl PageFlags {
...
+impl fmt::Debug for IndexEntry {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        if self.is_empty() {
+            f.debug_struct("IndexEntry")
+                .field("is_empty", &self.is_empty())
+                .finish()
+        } else {
+            f.debug_struct("IndexEntry")
+                .field("is_empty", &self.is_empty())
+                .field("page_index", &self.page_index().unwrap())
```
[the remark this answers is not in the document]
Swiftb0y: yeah thats another topic. We currently have loads of magic constants which we just assume are always the same, but we never check it. We should set up a system that can scan an entire database and see if any of our current assumptions over hardcoded values are violated, that would likely also greatly help with figuring out any patterns and the logic behind them. 
That said, in order for that to work, we shouldn't make it fatal in any way since that makes actually detecting these cases difficult. Instead, we should add a `Validation` trait that each type can implement to tests its own assumptions and flag them if requested. Its important to me though that these are **not** fatal. We should try to parse a DB if possible. This also means adding some more robust handling to some fields, though I haven't figured that you yet. I don't want to throw `#[br(try)]` at every member because a struct full of `Option`s gets really annoying.

src/pdb/mod.rs
Swiftb0y: yeah, I'd say avoid `as` in general if reasonable. It just silently truncates everything, which is seldom what you really want. https://doc.rust-lang.org/reference/expressions/operator-expr.html#r-expr.as.numeric

src/pdb/mod.rs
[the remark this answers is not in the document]
Swiftb0y: Ah we should probably add some way to stub out the pages we have not reverse-engineered yet so we can
1. create working re-exports
2. also create working exports with our own data that the CDJs don't choke on.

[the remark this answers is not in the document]
Swiftb0y: Ah yeah, and once we have Proxy API, we should add C/C++ bindings using [`cxx`](https://crates.io/crates/cxx) so we can integrate it into Mixxx or https://github.com/xsco/libdjinterop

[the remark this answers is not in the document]
Swiftb0y: created a couple issues to document this

src/pdb/mod.rs
```diff
@@ -129,6 +145,18 @@ pub enum PlainPageType {
 #[brw(little)]
 pub struct PageIndex(u32);
 
+impl TryFrom<u32> for PageIndex {
+    type Error = PdbError;
+
+    fn try_from(value: u32) -> Result<Self, Self::Error> {
+        if value <= 0x1FFF_FFFF {
+            Ok(Self(value))
```
Swiftb0y, line 153: `0x1FFF_FFFF` is the sentinel value, right? So its not a valid PageIndex?

src/pdb/mod.rs
```diff
@@ -441,13 +669,11 @@ impl RowGroup {
     pub fn present_rows(&self) -> &[Row] {
         &self.rows
     }
-    // TODO(Swiftb0y): Add a new error category for user APIs and add the correct
-    // error herer
-    #[allow(clippy::result_unit_err)]
+
```
Swiftb0y, line 672: thanks ;)

src/pdb/mod.rs
acrilique: You are right. Actually, if we consider the 3 bit shift we do to an IndexEntry to obtain its PageIndex, `0x1FFF_FFF` becomes `0x03FF_FFFF` which means the highest possible value for a PageIndex is `0x03FF_FFFE` (67108862 in decimal, so highest possible size for a database file is below 275GB if I'm not mistaken).

Swiftb0y: yeah definitely weird, but I don't think we can easily create a database file big enough to see what really happens then. 

Regardless, we should ensure PageIndicies are always in a valid range, so I guess check for `< 0x03FF_FFFF` then?

acrilique: Done 95d085c

PR #218, opened 2026-02-13
JSON output support for dump commands

stella3d:
when using the `dump-*` commands to inspect files, i found the `Debug` trait output a bit difficult to work with, so this adds JSON support to make it easier to read & inspect with tools like `jq`.

this is a bit opinionated in serializing byte slices as hex strings rather then arrays of `Number`, but i think this makes the output significantly easier to read & browse though.

some manual implementations of `Serialize`, like for `WaveformPreviewColumn`, are because the default serialization resulted in a hard to read format.

as an example, a settings dump in json format:

```json
{
  "brand": "PIONEER",
  "software": "rekordbox",
  "version": "0.001",
  "data": {
    "MySetting": {
      "unknown1": "0x7856341202000000",
      "on_air_display": "On",
      "lcd_brightness": "Two",
      "quantize": "On",
      "auto_cue_level": "Memory",
      "language": "English",
      "unknown2": 1,
      "jog_ring_brightness": "Bright",
      "jog_ring_indicator": "On",
      "slip_flashing": "On",
      "unknown3": "0x010101",
      "disc_slot_illumination": "Bright",
      "eject_lock": "Unlock",
      "sync": "Off",
      "play_mode": "Continue",
      "quantize_beat_value": "FullBeat",
      "hotcue_autoload": "On",
      "hotcue_color": "Off",
      "unknown4": 0,
      "needle_lock": "Lock",
      "unknown5": 0,
      "time_mode": "Remain",
      "jog_mode": "Vinyl",
      "auto_cue": "Off",
      "master_tempo": "On",
      "tempo_range": "TenPercent",
      "phase_meter": "Type2",
      "unknown6": 0
    }
  }
}
```

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -310,10 +310,7 @@ fn dump_anlz(path: &Path, format: DumpFormat) -> rekordcrate::Result<()> {
     let anlz = ANLZ::read(&mut reader)?;
     match format {
         #[cfg(feature = "json")]
-        DumpFormat::Json => println!(
-            "{}",
-            serde_json::to_string_pretty(&anlz).expect("failed to serialize ANLZ")
-        ),
+        DumpFormat::Json => println!("{}", serde_json::to_string_pretty(&anlz)?),
         DumpFormat::Debug => println!("{:#?}", anlz),
         #[cfg(not(feature = "json"))]
         _ => unreachable!("DumpFormat has no other variants without the json feature"),
@@ -368,10 +365,7 @@ fn dump_pdb(
             });
         }
         let dump = PdbDump { header, tables };
-        println!(
-            "{}",
-            serde_json::to_string_pretty(&dump).expect("failed to serialize PDB output")
-        );
+        println!("{}", serde_json::to_string_pretty(&dump)?);
         return Ok(());
     }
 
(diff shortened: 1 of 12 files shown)


src/main.rs
```diff
@@ -194,57 +223,99 @@ fn list_settings(path: &Path) -> rekordcrate::Result<()> {
...
-                    for (_, row) in data_content.rows {
-                        println!("      {:?}", row);
-                    }
+    match format {
+        DumpFormat::Json => {
+            let mut tables = Vec::new();
+            for table in &header.tables {
+                let mut pages = Vec::new();
+                for page in header
+                    .read_pages(
```
Swiftb0y: consider splitting match arms into separate functions?

Swiftb0y (commented):
this looks good (sorry for not taking a look at this sooner). One primary pain point for me is that serde is quite a heavy dependency and is only used for the json serialization. So I'd definitely prefer if the json serialization was an opt-in feature (since its only useful for the CLI tool, but not when rekordcrate is ever used as a library). Also, since this only about json serialization, using a leaner json library would also be preferred (though not a requirement if this features was optional).

acrilique:
Hi @Swiftb0y regarding the comment you left a couple months ago, I actually see that `serde` is already being used in the project, specifically for the XML-related stuff (and `quick-xml` also pulls in `serde` if I'm not mistaken). 
So in this case I'd rather use `serde-json` (ofc as opt-in, and potentially make xml opt-in too in another PR)

acrilique:
Hi @stella3d @Swiftb0y I rebased this one on top of main fixing conflicts (as since OP posted this we had the OffsetArray refactor, the LazyDatabase merge, anlz's LenPrefixedWideString merge and a few changes in main.rs). I would appreciate if any of you could take a look and maybe test it before I proceed with the merge.

acrilique (approved):

PR #233, opened 2026-04-07
feat(xml): publicise xml structs

radstevee:
also fixes a deserialize error when reading an empty playlist


Holzhaus (approved):
LGTM, thanks!

PR #239, opened 2026-05-27
feat(anlz): Add Support for 3-Band Waveforms

FizzyApple12:
This PR does the following:
1. Adds ``Waveform3BandPreview``, ``Waveform3BandPreviewColumn``, ``Waveform3BandDetail``, and ``Waveform3BandDetailColumn`` structs based on the information [here](https://djl-analysis.deepsymmetry.org/rekordbox-export-analysis/anlz.html#three-band-preview) and [here](https://djl-analysis.deepsymmetry.org/rekordbox-export-analysis/anlz.html#three-band-detail)
2. Adds ``Waveform3BandPreview`` (magic ``PWV6``) and ``Waveform3BandDetail`` (magic ``PWV7``) to ``ContentKind``
3. Adds``Waveform3BandPreview`` and ``Waveform3BandDetail`` to ``Content``
4. Cleans up a the documentation around the Waveforms to make the comments point to the correct links and have consistent descriptions


flake.nix
```diff
@@ -0,0 +1,41 @@
...
+          shellHook = ''
+            export PATH=$PATH:''${CARGO_HOME:-~/.cargo}/bin
+            export PATH=$PATH:''${RUSTUP_HOME:-~/.rustup}/toolchains/$RUSTC_VERSION-x86_64-unknown-linux-gnu/bin/
+          '';
+
+          LD_LIBRARY_PATH = pkgs.lib.makeLibraryPath nativeBuildInputs;
+        };
+      }
+    );
+}
```
acrilique: I think adding a nix flake is something that should go in another PR because it's an entirely different discussion

src/anlz.rs
```diff
@@ -440,6 +459,38 @@ pub struct WaveformColorDetailColumn {
...
+#[brw(big)]
+pub struct Waveform3BandDetailColumn {
+    /// Sound energy in the mid of the frequency range.
+    pub energy_mid_third_freq: u8,
+    /// Sound energy in the top of the frequency range.
+    pub energy_top_third_freq: u8,
+    /// Sound energy in the bottom third of the frequency range.
+    pub energy_bottom_third_freq: u8,
+}
+
```
acrilique, line 493: Aren't these the same? Why have 2 equal structs?

acrilique (changes requested):
Hi! I just left a couple of comments. Feel free to address them while you wait for a review by someone with write access.

flake.nix
FizzyApple12: Understandable, my original thought process was that adding this doesn't constitute a major change since it doesn't affect the rest of the project, but I'll move it to a new PR

src/anlz.rs
FizzyApple12: My reasoning was they originate from two different parts of the data and so keeping them separate like the rest of the waveform structs is consistent and gives flexibility to change one or the other if it's found that one or the other isn't formed correctly or has other deserialization problems that haven't been discovered

FizzyApple12: If you'd like, I can use the same struct for both though

flake.nix
FizzyApple12: Moved to #241

src/anlz.rs
acrilique: Nah it does feel consistent with the rest of the file. Nevermind

Swiftb0y:
It would be nice to have some tests for these to ensure stability. We already have some `.2EX` files in the complete_export dataset.

FizzyApple12:
The existing parsing tests that exist for anlz pass with this code, if there's interest I can make some more that test parsing individual parts of the data in the files though

Swiftb0y:
Ah right, anlz has no unit tests, only a basic read-write integration test.

Swiftb0y (approved):
LGTM. Thank you.

FizzyApple12:
Of course! I'm glad to contribute!

acrilique (approved):

Issue #240, opened 2026-05-27
ANLZ::read() Fails to Parse ExtendedCue Section of .EXT File


FizzyApple12:
When writing an application that needs access to the waveform and cue data in the .EXT analysis files, I found that EXT files on my rekordbox USB don't load correctly with an assertion failing in the parser:
```
failed to parse anlz file: 
 ╺━━━━━━━━━━━━━━━━━━━━┅ Backtrace ┅━━━━━━━━━━━━━━━━━━━━╸

 0: Error: no variants matched at 0xe7eb...
   ╭───────────────────────┄ BeatGrid ┄────────────────────┄
   ┆
   ┆assertion failed: `header.kind == ContentKind :: BeatGrid` at 0xe7eb
   ╰───────────────────────────────────────────────────────┄
   ╭───────────────────────┄ CueList ┄────────────────────┄
   ┆
   ┆assertion failed: `header.kind == ContentKind :: CueList` at 0xe7eb
   ╰──────────────────────────────────────────────────────┄
   ╭───────────────────────┄ ExtendedCueList ┄────────────────────┄
   ┆
   ┆ 0: Error: assertion failed: `(comment.len() as u32 + 1) * 2 == len_comment` at 0xe7f3
   ┆           While parsing field 'cues' in ExtendedCueList
   ┆     at src/anlz.rs:718
   ┆ 1: While parsing field 'self_0' in Content::ExtendedCueList
   ┆     at src/anlz.rs:596
   ┆
   ╰──────────────────────────────────────────────────────────────┄
   ╭───────────────────────┄ Path ┄────────────────────┄
   ┆
   ┆assertion failed: `header.kind == ContentKind :: Path` at 0xe7eb
   ╰───────────────────────────────────────────────────┄
   ╭───────────────────────┄ VBR ┄────────────────────┄
   ┆
   ┆assertion failed: `header.kind == ContentKind :: VBR` at 0xe7eb
   ╰──────────────────────────────────────────────────┄
   ╭───────────────────────┄ WaveformPreview ┄────────────────────┄
   ┆
   ┆assertion failed: `header.kind == ContentKind :: WaveformPreview` at 0xe7eb
   ╰──────────────────────────────────────────────────────────────┄
   ╭───────────────────────┄ TinyWaveformPreview ┄────────────────────┄
   ┆
   ┆assertion failed: `header.kind == ContentKind :: TinyWaveformPreview` at 0xe7eb
[...]

PR #242, opened 2026-06-01
fix(anlz): change type of `comment` field and capture trailing bytes

acrilique:
Addresses #240

`crate-digger` interprets the comment field as a fixed-length UTF-16BE string of exactly `len_comment` bytes, but `NullWideString` scans for a NUL terminator, potentially reading more bytes. `FixedUtf16String` reads the exact byte count and trims trailing NULs, aligning with the reference. Additionally, the `ExtendedCue` struct now captures any extra bytes beyond the defined fields as a `trailing` field.

Also fixed a typo (Rekordbot -> Rekordbox)

### DIFF
--- a/src/anlz.rs
+++ b/src/anlz.rs
@@ -224,6 +224,91 @@ pub struct Cue {
     unknown7: u32,
 }
 
+/// A fixed-length UTF-16BE string read from a specified number of bytes.
+/// Used for the `comment` field in the `ExtendedCue` section.
+///
+/// Unlike [`NullWideString`], this does not scan for a NUL terminator; it reads
+/// exactly `len_comment` bytes and decodes them. This matches the Kaitai spec's
+/// `size: len_comment` approach.
+#[derive(Debug, Default, PartialEq, Eq, Clone)]
+pub struct FixedUtf16String(pub String);
+
+impl FixedUtf16String {
+    /// Returns the number of bytes required to encode this string as UTF-16BE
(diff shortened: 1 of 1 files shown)


src/anlz.rs
```diff
@@ -264,13 +349,15 @@ pub struct ExtendedCue {
...
+    #[bw(calc = comment.byte_len())]
     len_comment: u32,
-    /// An UTF-16BE encoded string, followed by a trailing  `0x0000`.
-    #[br(assert((comment.len() as u32 + 1) * 2 == len_comment))]
-    pub comment: NullWideString,
+    /// An UTF-16BE encoded string, with a trailing NUL (`0x0000`).
+    ///
+    /// Parsed as a fixed-length field of `len_comment` bytes (per the Kaitai spec).
+    #[br(args(len_comment))]
+    pub comment: FixedUtf16String,
```
Swiftb0y, line 360: optionally, it may also make sense to fuse the length and the content into a single type the prefixed length field is an implementation detail.

Swiftb0y (approved):
LGTM. Thank you.

Holzhaus (approved):
Thank you!

For parsing bugs like these, it would be great if we have a test to make sure tha the fix is working as expected and that it doesn't break again. But I'm getting greedy 😅

acrilique:
I just addressed both of your suggestions (fusing `len_comment` and `comment` to a single type and adding a test) in 763e31b and 747e537 respectively

PR #248, opened 2026-06-06
feat(pdb): methods for pdb scaffolding

acrilique:
Here's part of the work I did on top of @RobinMcCorkell 's LazyDatabase PR (which was recently merged). I tried to clean it up a bit and select only enough to allow for proper review. It gets us one step closer to achieving serialization of the `pdb` file from scratch. I think Robin also has implemented something similar in his fork.

The tests are AI-generated and I haven't taken the time to check if they're actually providing value. Will surely do so before merging.

### DIFF
--- a/src/pdb/io.rs
+++ b/src/pdb/io.rs
@@ -87,6 +87,15 @@ pub struct Database<IO> {
     content: LazyDatabase,
 }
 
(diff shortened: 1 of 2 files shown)


acrilique:
closing due to #259

PR #249, opened 2026-06-07
Add render-waveforms CLI

RobinMcCorkell:
## Summary
- add a render-waveforms CLI and extract the SVG renderer into src/render_waveforms.rs
- extend ANLZ waveform parsing and tests, including PWVC support, waveform naming cleanup, and the PWV2 tiny-preview decoding fix
- switch SVG emission to xmlwriter for structured SVG output

## Validation
- cargo test --quiet

render

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -25,6 +25,7 @@ chrono = { version = "0.4", features = ["serde"], optional = true }
 rgb = { version = "0.8", optional = true }
 fallible-iterator = "0.3.0"
 
+xmlwriter = "0.1"
 [build-dependencies]
 glob = "0.3"
 
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -25,7 +25,6 @@ chrono = { version = "0.4", features = ["serde"], optional = true }
 rgb = { version = "0.8", optional = true }
 fallible-iterator = "0.3.0"
 
-xmlwriter = "0.1"
 [build-dependencies]
 glob = "0.3"
 
--- a/src/anlz.rs
+++ b/src/anlz.rs
@@ -68,26 +68,26 @@ pub enum ContentKind {
     VBR,
     /// Fixed-width monochrome preview of the track waveform.
     #[brw(magic = b"PWAV")]
-    WaveformPreview,
+    WaveformBluePreview,
     /// Smaller version of the fixed-width monochrome preview of the track waveform (for the
     /// CDJ-900).
     #[brw(magic = b"PWV2")]
-    TinyWaveformPreview,
+    WaveformBlueTinyPreview,
     /// Variable-width large monochrome version of the track waveform.
     ///
     /// Used in `.EXT` files.
     #[brw(magic = b"PWV3")]
-    WaveformDetail,
+    WaveformBlueDetail,
     /// Fixed-width colored preview of the track waveform.
     ///
     /// Used in `.EXT` files.
     #[brw(magic = b"PWV4")]
-    WaveformColorPreview,
+    WaveformRGBPreview,
     /// Variable-width large colored version of the track waveform.
     ///
     /// Used in `.EXT` files.
     #[brw(magic = b"PWV5")]
-    WaveformColorDetail,
+    WaveformRGBDetail,
     /// Fixed-width 3-band preview of the track waveform.
     ///
     /// Used in `.2EX` files.
@@ -98,6 +98,11 @@ pub enum ContentKind {
     /// Used in `.2EX` files.
     #[brw(magic = b"PWV7")]
     Waveform3BandDetail,
+    /// Per-band gain calibration for the 3-band player waveform.
+    ///
+    /// Used in `.2EX` files.
+    #[brw(magic = b"PWVC")]
+    Waveform3BandCalibration,
     /// Describes the structure of a sond (Intro, Chrous, Verse, etc.).
     ///
     /// Used in `.EXT` files.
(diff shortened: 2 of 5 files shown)


acrilique:
Hi Robin, thanks a lot! I haven't digged much into anlz yet so it's very convenient. Haven't read the changes yet but at first glance it seems weird to have a commit that touches the changelog. No one should ever have the need to manually edit the changelog, as it's auto-generated from the commit summaries that follow the conventions of the project.

RobinMcCorkell:
That was an over-zealous Copilot agent trying to update the changelog. I've tidied up the branch.

acrilique:
Cool, thanks. Something else that I'd be very grateful for is if you would use conventional commit summary format (as specified in the contribution guide) so the git hook is able to automatically update the changelog. In your last PR, I had to edit the commit summaries myself because of this.

To facilitate that task and also automatically run clippy and other checks at commit time, you might want to install the pre-commit hooks (again, this is explained in the contribution guide).

Again, thanks for the contribution and I'll get in touch when I'm able to properly review the code.

Cargo.toml
```diff
@@ -22,6 +22,7 @@ quick-xml = { version = "0.40.1", features = ["serialize", "serde-types"] }
 serde = { version = "1.0", features = ["derive"] }
 chrono = "0.4"
 fallible-iterator = "0.3.0"
+xmlwriter = "0.1"
```
acrilique: I see no need for this as we already have quick-xml as a dependency and it can achieve the same task with code that is very similar

src/main.rs
```diff
@@ -56,6 +58,24 @@ enum Commands {
...
+        path: PathBuf,
+        /// Output SVG file to write.
+        #[arg(value_name = "SVG_FILE")]
+        output: PathBuf,
+        /// Skip reading the sibling `.EXT` file.
+        #[arg(long)]
+        no_ext: bool,
+        /// Skip reading the sibling `.2EX` file.
+        #[arg(long)]
+        no_2ex: bool,
```
acrilique, line 78: Maybe for completeness we could add a `no_dat` argument?

src/anlz.rs
```diff
@@ -68,26 +68,26 @@ pub enum ContentKind {
...
     #[brw(magic = b"PWV4")]
-    WaveformColorPreview,
+    WaveformRGBPreview,
     /// Variable-width large colored version of the track waveform.
     ///
     /// Used in `.EXT` files.
     #[brw(magic = b"PWV5")]
-    WaveformColorDetail,
+    WaveformRGBDetail,
     /// Fixed-width 3-band preview of the track waveform.
```
acrilique, line 91: Why change the names? The previous ones seem better to me. And also for me it makes it a bit harder to find the actual code changes in the PR, so reviewing is harder

src/anlz.rs
```diff
@@ -470,96 +475,228 @@ pub struct ExtendedCue {
...
+    ///
+    /// This varies across the full `0..=127` range and seems related to overall preview intensity,
+    /// but its exact meaning is still unknown.
+    pub unknown3: u8,
+    /// Red channel intensity, corresponding to bass.
+    pub red_bass: u8,
+    /// Green channel intensity, corresponding to mids.
+    pub green_mids: u8,
+    /// Blue channel intensity, corresponding to highs.
+    pub blue_highs: u8,
```
acrilique: Also here, why put the color (which is only related to the rendering, so it's only tied to the cli app and not to the library) in the name? I think it is not an improvement over what we previously had

src/anlz.rs
```diff
@@ -470,96 +475,228 @@ pub struct ExtendedCue {
...
+
+    /// Coarse 5-bit column height.
+    pub const fn height(&self) -> u8 {
+        self.height
+    }
+
+    /// Raw low-order fine-height bits in on-disk order.
+    pub const fn low_bits(&self) -> u8 {
+        self.low_bits
+    }
```
acrilique: Here you also have the chance to make the PR diff smaller by just sticking to the previous convention, i.e. making the struct fields public

src/anlz.rs
```diff
@@ -570,12 +707,12 @@ pub struct WaveformColorDetailColumn {
...
-    /// Sound energy in the top of the frequency range.
-    pub energy_top_third_freq: u8,
-    /// Sound energy in the bottom third of the frequency range.
-    pub energy_bottom_third_freq: u8,
+    /// Low / bass band intensity.
+    pub low: u8,
+    /// Mid band intensity.
+    pub mid: u8,
+    /// High band intensity.
+    pub high: u8,
```
acrilique: According to the docs in the link above, "The three-band waveform preview entries are one-byte height values representing the mid-range, high, and low frequencies, in that order."

src/anlz.rs
```diff
@@ -586,12 +723,12 @@ pub struct Waveform3BandPreviewColumn {
...
```
acrilique: same for these

src/anlz.rs
```diff
@@ -1012,6 +1166,28 @@ pub struct Waveform3BandDetail {
...
+    pub unknown: u16,
+    /// Gain applied to the low / blue waveform band.
+    #[br(assert(header.content_size() == 6))]
+    pub low_gain: u16,
+    /// Gain applied to the mid / yellow waveform band.
+    pub mid_gain: u16,
+    /// Gain applied to the high / white waveform band.
+    pub high_gain: u16,
+}
+
```
acrilique, line 1132: This one isn't documented yet, right?

acrilique (changes requested):
Thank you so much! I left some comments and was able to test. I do see rendering to SVG as a cool feature for the `cli`

acrilique:
Hey @RobinMcCorkell , just pinging in case you missed my review.

acrilique:
I decided to address the issues I had with the PR myself and rebase, while also fixing clippy warnings and gating the `render-waveforms` command behind the `xml` feature, just as xml parsing now is also gated behind it. Will leave it here for a while

RobinMcCorkell:
Sorry I've been away recently.

Let me refactor this PR to reduce its overall diff size.

acrilique:
Hi @RobinMcCorkell could you explain a bit how did you get to implement `Waveform3BandCalibration`? What did you test and whether or not part of it is just speculation on your side?

PR #250, opened 2026-06-09
fix(pdb): align Ucs2le strings to 4 bytes

acrilique:
I had this lying around... My CDJ-350 players would instantly crash when reading a generated pdb export containing Ucs2le strings without this fix. Maybe someone has an intuition on a more general fix or better way to work around the issue.

### DIFF
--- a/src/pdb/offset_array.rs
+++ b/src/pdb/offset_array.rs
@@ -47,6 +47,20 @@ use super::PageHeapObject;
 use crate::util::{ArrayPolyfills, MaybeCalculated};
 use binrw::{binrw, io::SeekFrom, BinRead, BinResult, BinWrite};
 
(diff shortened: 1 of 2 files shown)


src/pdb/offset_array.rs
```diff
@@ -112,6 +126,20 @@ impl<T, const N: usize> OffsetArrayContainer<T, N> {
...
+        }
+
+        let align = u64::from(align);
+        let rem = offset % align;
+        if rem == 0 {
+            offset
+        } else {
+            offset + (align - rem)
+        }
+    }
```
Swiftb0y, line 132: isn't this just `offset.next_multiple_of(align)`?

acrilique: indeed! thanks

Swiftb0y (approved):

PR #259, opened 2026-07-16
Extending the `device` API to create exports from scratch

acrilique:
Until now, the API allowed users to read device exports. This PR restructures the `device` module and adds an (opinionated) high-level API to build a complete (and hopefully, working) export from scratch, as well as easy access to the underlying plain and ext databases via `device::Layout` methods when used alongside `pdb::io::Database::open`.

Example:

```rust
use rekordcrate::{DeviceExportWriter, Track};

let mut export = DeviceExportWriter::create("/mnt/usb")?;
let track = Track {
    title: "Song".into(),
    artist: "Artist".into(),
    file_path: "/Contents/Artist - Song.mp3".into(),
    ..Track::default()
};
export.add_track(&track)?;
export.close()?;
```

`add_track(&Track) -> AddTrackOutcome` resolves artists/albums/genres/keys/labels into deduplicated PDB rows. It returns the existing track id if `file_path` already exists, whether it was added earlier this session or read back by `open()`. 

Playlists (`create_playlist_folder`, `create_playlist`, `add_track_to_playlist`) enforce folder-vs-playlist roles and reject ids that don't resolve to an existing node or track.

Tag categories and track tags are also supported via `create_tag_category` and `add_tags_to_track`.

`Track` is a new struct that contains nothing of the inner intricacies of the actual `pdb` format: it has plain `String` fields, `rating` is an enum, `tempo` is a float, key-names are converted to a common format to avoid duplicates... A special quirk that I haven't mentioned before is that my CDJ-350 player would crash when entering the "TRACK" menu if there was any rows shorter than a specific number of bytes. After trial and error, I settled on a minimum of 221 bytes per track row (quite random, I know) and so the writer pads the `comment` field to achieve that minimum size.

`Artwork` is behind the artwork feature (given that it adds the `image` dependency): `add_track` copies from `artwork_path` into PIONEER/Artwork/ under under Rekordbox's bucketed folder layout (20 per folder), resized to 80×80 + 240×240. Without the feature we do not copy, and thus the job is left to the user.

**Limitations**: this is append-only: we still have no high-level way to do in-place modifications (imo just rebuilding the entire database from scratch is cheap so many users will just do that anyway). We also do not copy files into their destinations: users are expected to copy the file first, then fill `file_path` with the destination. There's also no way to generate `anlz` files (sadly my CDJ-350 doesn't seem to read or generate those at all so I would have no way to test that... unless I buy more hardware hehe).

Regarding the low-level read/write/edit access, I also made `pdb` struct fields public to facilitate it.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3,6 +3,12 @@
 version = 4
 
 [[package]]
+name = "adler2"
+version = "2.0.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "320119579fcad9c21884f5c4861d16174d0e06250625266f50fe6898340abefa"
+
+[[package]]
 name = "aho-corasick"
 version = "1.1.3"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -107,6 +113,12 @@ dependencies = [
 ]
 
 [[package]]
+name = "bitflags"
+version = "2.13.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "b4388bee8683e3d04af747c73422af53102d2bd24d9eadb6cbc100baef4b43f8"
+
+[[package]]
 name = "bumpalo"
 version = "3.19.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -117,6 +129,12 @@ name = "bytemuck"
 version = "1.24.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "1fbdf580320f38b612e485521afda1ee26d10cc9884efaaa750d383e13e3c5f4"
+
+[[package]]
+name = "byteorder-lite"
+version = "0.1.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "8f1fe948ff07f4bd06c30984e69f5b4899c516a3ef74f34df92a2df2ab535495"
 
 [[package]]
 name = "cc"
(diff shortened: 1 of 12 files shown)


acrilique:
Also, sorry about the size of this one...

sabjorn:
I am so friggin' stoked for this

Swiftb0y:
> Artwork is behind the artwork feature (given that it adds the image dependency): add_track copies from artwork_path into PIONEER/Artwork/ under under Rekordbox's bucketed folder layout (20 per folder), resized to 80×80 + 240×240. Without the feature we do not copy, and thus the job is left to the user.

fwiw this is also not required in order to load exports. Creating an export in rekordbox and deleting the artwork folder from the USB still works without any issues on the CDJs I've tested. 

> this is append-only: we still have no high-level way to do in-place modifications (imo just rebuilding the entire database from scratch is cheap so many users will just do that anyway)

This is also absolutely fine IMO :+1:

acrilique:
> fwiw this is also not required in order to load exports. Creating an export in rekordbox and deleting the artwork folder from the USB still works without any issues on the CDJs I've tested.

It's definitely not required, but IMO we should aim to provide users with the ability to produce exports that use most (if not all) of the features in the pdb format (while abstracting away the intricacies and still allowing low-level access). That's why I added it; do you agree with that and also with the pattern chosen (gate functionality behind a feature, as it requires a new dependency)?

PR #268, opened 2026-08-29
fix(anlz): Correct waveform data structure bugs and add missing PWVC section

RobinMcCorkell:
This is a focused revival of #249, limited to the `anlz` waveform data structure fixes and their tests.

- Fixes `TinyWaveformPreviewColumn` so the height is read from the upper nibble as seen in real files.
- Decodes `WaveformColorDetailColumn` manually and exposes the two fine-height bits instead of treating them as unknown.
- Adds support for the `PWVC` 3-band waveform calibration section.
- Adds Clone/Copy derives needed by waveform consumers.
- Adds unit tests for calibration round-trips, 3-band section round-trips, and RGB detail bit layout.

Closes #249.

### DIFF
--- a/src/anlz.rs
+++ b/src/anlz.rs
@@ -111,6 +111,11 @@ pub enum ContentKind {
     /// Used in `.2EX` files.
     #[brw(magic = b"PWV7")]
     Waveform3BandDetail,
+    /// Per-band gain calibration for the 3-band player waveform.
+    ///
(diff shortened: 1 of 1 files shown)


acrilique (approved):
Thanks a lot for doing this Robin, this one is much easier to just merge right away. LGTM as is.

PR #269, opened 2026-08-29
feat(device_library_plus): Add Device Library Plus support

RobinMcCorkell:
This PR adds **read-only MVP support** for Rekordbox Device Library Plus (`exportLibrary.db`) databases: encrypted SQLite databases used by Pioneer Rekordbox.

The implementation is gated behind the `device_library_plus` Cargo feature and includes:

- Bundled SQLCipher support and Rekordbox’s default database key, allowing encrypted `exportLibrary.db` files to be opened.
- A Diesel SQLite schema and read-only models for all mirrored Device Library Plus tables.
- `TableRecord::all` helpers used to load and display every supported table.
- A `dump-dlp <DLP_FILE>` CLI command that prints each supported table and its rows.
- Diesel connection and query errors integrated into `RekordcrateError`.
- The mirrored SQL schema (`src/device_library_plus/schema.sql`) as a reference for the database layout.
- A bundled encrypted `exportLibrary.db` fixture and an integration test that decrypts it and loads property/content rows through the same table-loading API used by `dump-dlp`.

Deliberately out of scope for this MVP: database creation, row insertion or updates, default seed data, record constructors/builders, and arbitrary SQLCipher keys.

**Copyright**: The models were generated by AI after reading the sqlite database columns and are not derived from other projects of a similar nature. The code implementation is similarly original, deriving only from other code in this repository.


acrilique:
This is great Robin. I already saw an attempt from yours in your fork but I really appreciate you doing the effort to post it as a PR. Before I review I wanted to ask, did you base this work on any other project (like pyrekordbox or rbox or anything like that)? Attribution and ensuring license compatibility would be needed in that case.

RobinMcCorkell:
> This is great Robin. I already saw an attempt from yours in your fork but I really appreciate you doing the effort to post it as a PR. Before I review I wanted to ask, did you base this work on any other project (like pyrekordbox or rbox or anything like that)? Attribution and ensuring license compatibility would be needed in that case.

I added a note in the PR description: this code is original and only the sqlite database itself (i.e. table schemas) and other code in this repository were used to derive this work.

PR #273, opened 2026-08-29
feat(pdb): type opaque table 7 and 18 rows

RobinMcCorkell:
These two tables appear to be essential to creating a functioning PDB file that will be read by Pioneer equipment, so we need a way to construct them in code even though we don't know how they work exactly.

## Summary

- add typed `Unknown18Row` support for the fixed 8-byte rows in plain PDB table 18
- add typed `Unknown7Row` support for the fixed 60-byte rows in `exportExt.pdb` table 7
- model table-7 flag words as preserving bitfields and integrate both row types with `Row`, `RowVariant`, and page-heap sizing
- cover parsing against the existing demo export fixtures and direct row/enum serialization round trips

This is intentionally limited to the two previously opaque, fixed-layout tables; other unknown page types remain opaque.

## Checks

- `cargo fmt -- --check`
- `cargo clippy --all-targets --all-features -- -D warnings`
- `cargo check --no-default-features --all-targets`
- `cargo test --all-targets --all-features`

### DIFF
--- a/src/pdb/bitfields.rs
+++ b/src/pdb/bitfields.rs
@@ -16,7 +16,7 @@ use modular_bitfield::prelude::*;
 #[cfg(feature = "json")]
 use serde::{Serialize, Serializer};
 
-use crate::pdb::RowGroup;
+use crate::pdb::{PageHeapObject, RowGroup};
 
 /// Packed field found in the page header containing:
 /// - number of used row offsets in the page (13 bits).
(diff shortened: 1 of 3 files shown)


acrilique:
I've been able to generate working exports while leaving these two tables completely empty. I'm not 100% against this tho, as it allows easier inspection by users of the bytes inside those tables. I'll try to make some time to review

PR #276, opened 2026-09-01
fix(anlz): correct 3-band waveform field order, add sweep fixture

RobinMcCorkell:
## Summary

Fixes the frequency-band field ordering of the 3-band waveform columns, verified against a new test fixture generated from a real frequency sweep:

- `Waveform3BandPreviewColumn` (PWV6) and `Waveform3BandDetailColumn` (PWV7) documented the bytes as (mid, high, low) — following the [djl-analysis docs](https://djl-analysis.deepsymmetry.org/rekordbox-export-analysis/anlz.html#three-band-preview) — but the actual byte order is **(low, mid, high)**.
- Adds `data/anlz/sweep/`: ANLZ analysis files of a track containing a 20 kHz→20 Hz sweep repeated three times (stereo, left channel, right channel). Because the dominant band at any point of the track is known, the fixture pinpoints which byte encodes which band.
- Adds a regression test asserting the band energies at 5% (high), 15% (mid), and 25% (low) of the track duration.

Note that the [djl-analysis documentation](https://djl-analysis.deepsymmetry.org/rekordbox-export-analysis/anlz.html#three-band-preview) itself may need updating (a PR is open at Deep-Symmetry/crate-digger#48 for crate-digger's copy of this documentation (refs Deep-Symmetry/crate-digger#22)).

## Validation

- `cargo test`
- `cargo clippy --all-targets --all-features -- -D warnings`

## Related

- The waveform rendering PR #275 builds on this fix and renders color previews using the corrected band mapping. The combined stacked branch (all preview formats) is drafted as RobinMcCorkell/rekordcrate#2.


FizzyApple12:
I'm glad that you ran in to this too and decided to actually test it. When I was experimenting a few months ago with rendering them, I assumed something was wrong in my rendering code and ended up doing this exact same swap to make it look more accurate.

acrilique:
Hello! Thanks a lot for the contribution Robin. Indeed, this is the actual order, and given the change is quite self-contained, I will merge right away. Also thanks for also sending a PR to `crate-digger`.

FYI @RobinMcCorkell @FizzyApple12 (and anyone else interested) I have been doing some reverse engineering of the rekordbox binary via static analysis with ghidra, and I've got a good approximation of what DSP is rekordbox doing to generate the various ANLZ sections (i.e. I know more about the exact filters, and meaning of previous unknowns). I'm eager to publish but given that I was very AI assisted I want to curate the info first. 

Also Robin please have a bit of patience for me to review all your recent PRs as I've just gotten to a new job.

FizzyApple12:
@acrilique I'd still be very excited to read about your findings! Whenever you publish them, please send a link!

acrilique:
@FizzyApple12 I released [rekordlib](https://github.com/acrilique/rekordlib). For info on the meaning of waveform-related anlz sections, see [rekordlib.anlz.Content](https://acrilique.github.io/rekordlib/#rekordlib.anlz.Content). For the reverse-engineered analysis functions, see [rekordlib.anlz.buildColumnsFromPcm](https://acrilique.github.io/rekordlib/#rekordlib.anlz.buildColumnsFromPcm).

Let me know what you think. As I say in the README and USAGE of the repo, that library was kind of a summer experiment to me and I actually intend to bring back the obtained knowledge to rekordcrate.

FizzyApple12:
@acrilique This is insanely cool and I'm very excited to see that you're planning to bring these changes back to rekordcrate! The code looks very sensible and is reasonable to follow and I think having a higher level API for interacting with the library is a great idea! I'd maybe expand some of the variable names used in the analysis algorithms when back-porting changes for clarity sake but that's mostly a nitpick lol.

I'm guessing that you use those frequency test files in analysis tests to narrow down the frequencies on the biquad filters for the waveform generator? Did you happen to stumble upon the filter configurations used in their DSP pipeline that produces master/cue output? I'd personally be curious to see how they match up with the waveform generation filters. Also, the scales on the 3-Band is very intriguing and I'm wondering why they're storing that when you directly mention the scaling is already applied.

Regardless, thank you for the excellent work and great reference material!

acrilique:
Thanks for the nice words @FizzyApple12 . 

I'm not 100% sure we do want the analysis feature in rekordcrate as it's more code to maintain and technically a user program can do DSP itself... I'm still thinking of an API to provide ease of use and flexibility at the same time for anlz generation.

Indeed, tailored testfiles were used for things like narrowing down frequencies of filters. I also told the AI to use ghidra's static analysis with the rekordbox 6.8.5 executable to fill other gaps. I'm currently not aware of all the details, but if you're interested in anything specific I can dig into my conversation history and the folder where the AI dumped all the files of the reverse engineering process.

> Did you happen to stumble upon the filter configurations used in their DSP pipeline that produces master/cue output?

Do you mean the filters for the DJing functionality of rekordbox? I unfortunately didn't look for those, but it would indeed be interesting to see what exact functions are they using for DSP.

The scales are also a mystery to me. Maybe the cdj players do the opposite operation to get the original value? Not sure for what tho

